PR #240, opened 2026-03-17
feat(claude): host-side marketplace cloning for private plugin repos

abezzub-dr:
## Summary
- Clone marketplace repos on the **host** (where git credentials exist) instead of inside the Docker build context
- Copy marketplace files into the build context and write `known_marketplaces.json` directly
- `claude plugin install` finds plugins from the local filesystem — no network needed during build
- Falls back gracefully to build-time `claude plugin marketplace add` if host clone fails

This enables private plugin marketplaces (e.g., company-internal GitHub repos) to work in moat containers without leaking credentials into the Docker image.

## Changes
1. `internal/providers/claude/marketplace.go` — `CloneMarketplace`, `CollectMarketplaceFiles`, `GenerateKnownMarketplaces`
2. `internal/providers/claude/dockerfile.go` — Handle `PreCloned` marketplaces with COPY + known_marketplaces.json
3. `internal/deps/dockerfile.go` — Pass through `ExtraContextFiles` to build context
4. `internal/run/manager.go` — Clone marketplaces before image build, merge into context

## Test plan
- [x] Unit tests pass: `go test ./internal/providers/claude/ -v`
- [x] Existing deps tests pass: `go test ./internal/deps/ -v`
- [x] Configure a private marketplace in moat.yaml and verify plugins install


dpup (commented):
The design is sound and solves the real problem (private repos need host credentials, not build-time credentials). But there's one P1 that blocks merge and several P2s worth fixing.

## 🔴 P1 — `Ref` field is silently ignored (blocks merge)

`config.MarketplaceSpec` has a `Ref string` field with a config test that asserts it round-trips. `CloneMarketplace` accepts a `ref` parameter and passes it as `--branch`. But `manager.go:1258` hardcodes it to `""`:

```go
clonedDir, cloneErr := claude.CloneMarketplace(ctx, m.Repo, "")
```

A user who sets `ref: v2.0` in `moat.yaml` silently gets the default branch. Additionally, `Ref` isn't included in the image hash, so changing the ref wouldn't even trigger a rebuild.

Either wire it through (propagate `Ref` from config → `MarketplaceConfig` → `CloneMarketplace`, and include it in `ImageTag`), or remove the field from the schema until it's implemented. The current state is a documented API that does nothing.

## 🟡 P2 — Should fix before merge

**`time.Now()` in `GenerateKnownMarketplaces`** (`marketplace.go:128`)

Every call produces a different `known-marketplaces.json`, invalidating Docker's layer cache on every build even when nothing changed. The rest of the system is carefully content-addressed — this breaks that invariant. Use a fixed constant (`"1970-01-01T00:00:00Z"`) or the repo's last commit timestamp.

**Clone loop runs before `ImageExists` check** (`manager.go:1256` vs `1359`)

The clone + collect loop runs unconditionally. If the image is already cached (the common case), all marketplace repos are cloned and all files are read into memory for nothing. Gate the clone loop on `!exists`, consistent with how the build step itself is gated.

**`ExtraContextFiles` inside `if ScriptName != ""`** (`deps/dockerfile.go:214–219`)

`known-marketplaces.json` is only merged into the build context inside the `ScriptName` guard. Today this always holds (pre-cloned marketplaces imply plugins imply a script), but the coupling is implicit. If a future caller sets `ExtraContextFiles` without a script, the files are silently dropped while the Dockerfile still has `COPY` directives referencing them — a confusing build failure. Move the `ExtraContextFiles` merge outside the condition.

**Validate marketplace names in `manager.go` before `CollectMarketplaceFiles`** (`manager.go:~1262`)

The name from `ExtraKnownMarketplaces` is used to construct build-context map keys (`"marketplaces/" + name + "/"`) before `validMarketplaceName` is checked. Validation happens later in `GenerateDockerfileSnippet`. Check `validMarketplaceName` at the top of the loop in `manager.go` and skip early on invalid names — don't let an invalid name influence the build-context map at all.

**Add `--no-recurse-submodules`** (`marketplace.go:86`)

`git clone --depth 1` doesn't prevent submodule initialization. A malicious marketplace with submodules could trigger hooks on the host. Add `--no-recurse-submodules` to the clone args.

## 🔵 P3 — Nice to have

- **`validPreClonedPath` allows `..`** — the regex `^[a-zA-Z0-9._/-]+$` permits traversal sequences. Not exploitable today (`PreCloned` is always `"marketplaces/" + validatedName`), but misleadingly permissive. Consider `^[a-zA-Z0-9_-]+(/[a-zA-Z0-9_.-]+)*$` or an explicit `strings.Contains(m.PreCloned, "..")` guard.
- **`fmt.Sprintf` with no format args** at `dockerfile.go:138` — plain string literal suffices.
- **`ref` parameter** — if not wiring it up now, remove it from `CloneMarketplace`'s signature until there's a concrete use case with a config field behind it.
- **File size guard in `CollectMarketplaceFiles`** — no bound on per-file or aggregate size. A large binary in a marketplace repo loads entirely into memory. Worth a size check with a `log.Warn` and skip.

dpup (commented):
Review comments on the host-side marketplace cloning logic.

dpup:

Issue #234, opened 2026-03-14
gh CLI broken by http_unix_socket: null in config


dpup:
## Problem

`gh` CLI fails with `dial unix null: connect: no such file or directory` when making API calls. This prevents `gh pr create`, `gh api`, and other commands from working.

## Symptoms

```
$ gh api user
Get "https://api.github.com/user": dial unix null: connect: no such file or directory

$ gh auth status
X Failed to log in to github.com using token (GH_TOKEN)
- The token in GH_TOKEN is invalid.
```

## Root cause (unconfirmed)

`~/.config/gh/config.yml` contains `http_unix_socket: null`. The YAML `null` value appears to be interpreted by `gh` as the literal string `"null"`, causing it to attempt connecting to a unix socket at path `null` instead of using normal HTTP.

**Hypothesis:** When `gh` detects an expired or invalid token, it rewrites `config.yml` and serializes the nil `http_unix_socket` field as YAML `null`. On subsequent runs, this gets deserialized as the string `"null"` rather than empty/nil.

## Workaround

Change `http_unix_socket: null` to `http_unix_socket: ""` in `~/.config/gh/config.yml`.

## Moat context

This was observed inside a Moat sandbox where GitHub credentials are injected at the proxy layer. The `GH_TOKEN` env var is set to a placeholder that gets replaced by the proxy — but the `gh` CLI never reaches the proxy because it fails on the unix socket dial first.

## Questions

- What triggers `gh` to rewrite `config.yml` with `null` values?
- Is this a `gh` CLI bug (string/nil confusion) or a Moat issue (something we do causes the config rewrite)?
- Should Moat's init-files for gh config prevent this?

PR #241, opened 2026-03-17
fix(github): strip nil values from gh config to prevent YAML null round-trip

abezzub-dr:
## Summary
Fixes `dial unix null: connect: no such file or directory` errors when using `gh` CLI inside moat containers.

## Root cause
The gh CLI config (`~/.config/gh/config.yml`) commonly has empty fields like:
```yaml
http_unix_socket:
editor:
pager:
```

Moat copies this config into the container after sanitizing credentials. The YAML round-trip (unmarshal → re-marshal) converts empty/nil values to the string `"null"`:
```yaml
http_unix_socket: null
```

The `gh` CLI interprets this as a Unix socket path and tries to dial it, producing the error.

## Fix
- Strip `http_unix_socket` entirely (proxy handles HTTP transport)
- Remove all nil values to prevent the same issue with other empty fields

## Test plan
- [x] `go test ./internal/providers/github/ -run TestSanitize -v`
- [x] Run `moat run --grant github -- gh api user` and verify it works

### DIFF
--- a/internal/providers/github/provider.go
+++ b/internal/providers/github/provider.go
@@ -105,11 +105,13 @@ func sanitizeGhConfig(content []byte) ([]byte, error) {
 	}
 	delete(cfg, "hosts")
 	delete(cfg, "oauth_token")
+	// Remove http_unix_socket — the proxy handles HTTP transport.
 	// Remove nil values to prevent YAML null serialization.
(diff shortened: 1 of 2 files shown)


dpup:
Related: #234

PR #250, opened 2026-03-17
fix(daemon): complete run setup before registry insertion to prevent proxy race

### LINKED ISSUE #195: Proxy returns 407 for fast-starting containers (registration race)
## Summary

When a container starts and immediately makes HTTPS requests through the proxy, `curl` receives `407 Proxy Authentication Required`. Adding a 2-3 second sleep before the first request makes it work reliably. The proxy daemon's `RegisterRun()` call returns successfully, but the daemon isn't fully ready to serve CONNECT tunnels for that run's auth token yet.

## Reproduction

```bash
# Fails — curl fires before proxy is ready
moat run --grant github -- sh -c 'curl -sS https://api.github.com/user'
# => curl: (56) CONNECT tunnel failed, response 407

# Works — 2 second delay gives proxy time to be ready
moat run --grant github -- sh -c 'sleep 2 && curl -sS https://api.github.com/user'
# => 200 OK, credential injected
```

## Observed behavior

- `manager.Create()` calls `daemon.RegisterRun()` which returns successfully
- Container starts via `StartAttached()` with `HTTP_PROXY` set to `http://moat:<token>@host:port`
- Container's first HTTPS request sends `CONNECT` with `Proxy-Authorization: Basic <token>`
- Proxy responds with `407 Proxy Authentication Required`
- Debug logs show zero proxy activity for the failed run (no `proxy connect`, no `credential injected`)
- `moat proxy status` shows `Active runs: 0` after the run completes

## Expected behavior

The proxy should be ready to serve requests for a run immediately after `RegisterRun()` returns.

## Environment

- macOS (Docker runtime)
- Proxy daemon running as background process via Unix socket API

## Analysis

The `RegisterRun()` RPC returns to the CLI process, which then starts the container. But there may be a window where the daemon has acknowledged the registration but hasn't yet updated its internal routing table that the proxy's `ServeHTTP` handler reads. This could be:

1. A lock contention issue between the registration handler and the proxy handler
2. The registration updating a data structure that the proxy reads asynchronously
3. A goroutine scheduling delay between the socket handler and the proxy's run lookup

The race is timing-sensitive — containers that start slowly (image build, large entrypoint) never hit it, but fast `sh -c` commands reliably fail.

dpup:
## Summary

- Fixes #195: Proxy returns 407 for fast-starting containers due to registration race
- Generates the auth token upfront (before registry insertion) so AWS credential provider and token refresh can be fully initialized before the `RunContext` becomes visible to the proxy
- Previously, `Register()` was called first, then AWS setup happened after — a fast-starting container could hit the proxy and find an incomplete `RunContext`, resulting in a 407 response

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/daemon/...` passes
- [x] `make lint` passes (0 issues)
- [ ] Manual test: run a container with AWS grants and verify no 407 on first request
- [ ] Manual test: verify daemon restart re-registration still works (AuthToken path)

### DIFF
--- a/internal/daemon/persist.go
+++ b/internal/daemon/persist.go
@@ -216,6 +216,8 @@ func RestoreRuns(ctx context.Context, registry *Registry, runs []PersistedRun) i
 
 		// Create a per-run context and set cancel BEFORE registering so that
 		// a concurrent handleUnregisterRun sees a valid cancel function.
+		// Set up AWS credentials BEFORE RegisterWithToken so the RunContext is
+		// fully initialized before the proxy can observe it.
 		// This matches the ordering in handleRegisterRun.
 		runCtx, cancel := context.WithCancel(ctx)
 		rc.SetRefreshCancel(cancel)
@@ -224,8 +226,6 @@ func RestoreRuns(ctx context.Context, registry *Registry, runs []PersistedRun) i
 		if len(pr.Grants) > 0 {
 			StartTokenRefresh(runCtx, rc, pr.Grants)
 		}
-
-		registry.RegisterWithToken(rc, pr.AuthToken)
 
 		// Set up AWS credential provider if configured.
 		if pr.AWSConfig != nil {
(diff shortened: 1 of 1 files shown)


dpup:
Fixed the same ordering issue in persist.go — AWS provider setup now happens before RegisterWithToken, matching server.go.

Issue #277, opened 2026-03-23
grant placeholders for anthropic and gemini lack format-valid prefixes


dpup:
## Problem

`grantToPlaceholder` in `internal/run/manager.go` returns the generic `credential.ProxyInjectedPlaceholder` (`"moat-proxy-injected"`) for `anthropic` and `gemini` grants. SDKs that validate key format locally will reject these before making any HTTP request, so the proxy never gets a chance to inject the real credential.

This is the same class of bug already fixed for:
- `github` → `GitHubTokenPlaceholder` (needs `ghp_` prefix)
- `openai` → `OpenAIAPIKeyPlaceholder` (needs `sk-` prefix)

## Impact

An MCP server using `grant: anthropic` or `grant: gemini` in `codex.mcp:` or `gemini.mcp:` would fail at SDK key validation before any network request is made.

## Fix

Add format-valid placeholder constants for `anthropic` (needs `sk-ant-api` or `sk-ant-oat` prefix) and `gemini`, and return them from `grantToPlaceholder`.

Found by the review bot on #184.

PR #293, opened 2026-03-30
fix: add Anthropic and Gemini token placeholders

daveworth:
As noted in #277 these two conditions were missing.  We've also covered this (internal) function in a test for completeness.


dpup (approved):
Thanks for the fix!

dpup:

PR #305, opened 2026-04-06
fix: install Rust toolchain to shared location for non-root access

joonas:
While working on a Rust project, I found that the model could not make use of the Rust toolchain even though I had `rust` declared as dependency.

This change moves the Rust tooling to be installed under `/usr/local/cargo`, makes it world readable and executable, and finally adds it to PATH.


dpup:

PR #306, opened 2026-04-07
feat(claude): add support for managed settings

abezzub-dr:
## Summary

Claude managed settings flow: if there are managed settings enforced by user's organization, Claude prompts user on startup and immediately exists regardless of whether user accepted or rejected the managed settings prompt. On the next startup Claude runs normally if settings were accepted. Moat does not store the result of settings prompt, so users sees managed settings prompt every time and it is impossible to start the Claude in moat.

Sample managed settings prompt:
```
 Managed settings require approval 

 Your organization has configured managed settings that could allow execution of arbitrary code or interception of your prompts and responses.                                                                              
   
 Settings requiring approval:                                                                                                                                                                                               
   · otelHeadersHelper                                                                                                                                                                                                    
   · OTEL_EXPORTER_OTLP_ENDPOINT
   · hooks

 Only accept if you trust your organization's IT administration and expect these settings to be configured.

 ❯ 1. Yes, I trust these settings
   2. No, exit Claude Code
```

Changes (written by Claude):
- Fixes a bug where organizations with server-managed settings (hooks, OTEL, etc.) saw Claude Code's "Managed settings require approval" prompt on **every** container startup, making it impossible to use moat with Claude.
- Root cause: Claude Code caches server-managed settings in `~/.claude/remote-settings.json` after user approval. On subsequent startups, it compares fetched settings against this cache — if they match, no prompt. Moat wasn't copying this cache into containers.
- Copies `remote-settings.json` from the host into the container staging directory and places it at `~/.claude/remote-settings.json` via `moat-init.sh`.

## Test plan

- [x] Run `moat run` with a Claude grant on an org that has server-managed settings — verify no approval prompt appears
- [ ] Run `moat run` without server-managed settings (no `~/.claude/remote-settings.json`) — verify no regression
- [x] Verify existing unit tests pass (`make test-unit`)

### DIFF
--- a/internal/deps/scripts/moat-init.sh
+++ b/internal/deps/scripts/moat-init.sh
@@ -86,6 +86,14 @@ if [ -n "$MOAT_CLAUDE_INIT" ] && [ -d "$MOAT_CLAUDE_INIT" ]; then
     chmod 600 "$TARGET_HOME/.claude/.credentials.json"
   fi
 
+  # Copy remote-settings.json if present (server-managed settings cache)
+  # This prevents Claude Code from prompting for managed settings approval
+  # on every container startup by providing the cached approval state.
+  if [ -f "$MOAT_CLAUDE_INIT/remote-settings.json" ]; then
+    cp -p "$MOAT_CLAUDE_INIT/remote-settings.json" "$TARGET_HOME/.claude/"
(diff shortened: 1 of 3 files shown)


dpup:

PR #308, opened 2026-04-08
fix(claude): enable 1M context detection inside moat containers

abezzub-dr:
## Summary

- Fix Claude Code not detecting 1M context window subscription inside moat containers
- Three issues combined: non-OAuth-looking placeholder token, missing bootstrap response, no subscription metadata in credentials
- Cache `/api/bootstrap` response at grant time (when host OAuth token is fresh) and serve it via proxy
- Use `sk-ant-oat01-*` placeholder in `.credentials.json` instead of `CLAUDE_CODE_OAUTH_TOKEN` env var
- Include subscription metadata in credentials file and synthetic profile responses

## Context

Claude Code calls `/api/bootstrap` at startup to determine account capabilities. Setup-tokens (from `claude setup-token`) lack the scopes for this endpoint to return proper account info and feature flags, causing Claude Code to report "Opus 4.6 with 1M context is not available for your account."

This PR caches the full bootstrap response at grant time (when the host's short-lived OAuth token is guaranteed fresh), stores it in the encrypted credential store, and has the proxy serve it instead of the limited setup-token response. As a fallback, it also attempts to fetch bootstrap at run time if the host token is still valid.

Additionally, the `CLAUDE_CODE_OAUTH_TOKEN` env var (set to `moat-proxy-injected`) is replaced with a proper `.credentials.json` containing an `sk-ant-oat01-*` prefix placeholder, so Claude Code recognizes the session as OAuth-authenticated.

## Test plan

- [x] Run `moat grant claude --verbose` and verify `cached bootstrap response at grant time` appears with non-zero `bootstrap_len`
- [x] Run `moat claude` and verify Claude Code reports 1M context availability
- [ ] Verify `moat claude` still works with `anthropic` (API key) grants (no regression)
- [x] Verify unit tests pass: `make test-unit`
- [x] Verify lint passes: `make lint`

### DIFF
--- a/docs/content/reference/04-grants.md
+++ b/docs/content/reference/04-grants.md
@@ -124,7 +124,7 @@ Stored as `anthropic.enc`.
 
 The proxy injects credentials for requests to `api.anthropic.com`:
 
-- **`claude` grant**: `Authorization: Bearer <token>` with OAuth beta flag. Container receives `CLAUDE_CODE_OAUTH_TOKEN` placeholder.
+- **`claude` grant**: `Authorization: Bearer <token>` with OAuth beta flag. Container receives a `.credentials.json` with an OAuth placeholder token and subscription metadata.
 - **`anthropic` grant**: `x-api-key: <key>`. Container receives `ANTHROPIC_API_KEY` placeholder.
 
 ### Refresh behavior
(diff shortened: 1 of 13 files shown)


dpup:
Do you mind if we split this up. The placeholder token and expiresAt seem simple an clearly correct. I'm a little more worried about the bootstrap caching and subscription stuff.

For the second part I think we should also orient around grant-time capture, run-time synthesize. I like the current implementation has cross `--profile` bleed.

abezzub-dr:
I don't have time to investigate this further, but I opened separate PRs for more obvious changes like you suggested.

PR #310, opened 2026-04-10
feat(gcloud): add Google Cloud credential provider

dpup:
## Summary

- Add `moat grant gcloud` provider that authenticates gcloud CLI and Google client libraries inside containers via GCE metadata server emulation
- Host daemon mints short-lived OAuth2 access tokens using Application Default Credentials and serves them through an emulated metadata endpoint — no long-lived credentials enter the container
- Support `--project`, `--key-file`, `--impersonate-service-account`, and `--scopes` CLI flags
- Wire into daemon persistence, proxy routing (metadata.google.internal + 169.254.169.254), and run manager
- Handle direct (non-proxied) metadata requests for gcloud CLI's GCE detection, which bypasses HTTP_PROXY via `GCE_METADATA_ROOT`
- Support multiple concurrent gcloud runs sharing the same credential profile
- Add guide, reference docs, and changelog entry

## Architecture

Follows the AWS provider pattern: host-side `CredentialProvider` with token caching → `EndpointHandler` serving GCE metadata routes → per-run handler attached via daemon `RunContext` → proxy intercepts metadata requests via `HTTP_PROXY`.

Key differences from header-injection providers (GitHub, Claude):
- No static token stored — tokens minted on demand from host ADC
- Metadata server emulation instead of header injection
- Token cached with 5-minute refresh buffer

### Direct metadata handling

gcloud CLI and Python's `google-auth` library use bare `http.client` for GCE metadata detection, bypassing `HTTP_PROXY`. The proxy handles these direct requests by:
1. Detecting `Metadata-Flavor: Google` header on non-proxied requests
2. Resolving the gcloud handler via the daemon registry (no auth token available)
3. Returning the handler only when all registered gcloud runs share the same credential profile — different profiles return 403 to prevent cross-run credential leakage

### Container environment

Sets `GCE_METADATA_HOST`, `GCE_METADATA_ROOT`, `CLOUDSDK_CORE_ACCOUNT`, `GOOGLE_CLOUD_PROJECT`, and `CLOUDSDK_CORE_PROJECT` to point the gcloud CLI and Google client libraries at the proxy's metadata emulator.

## Followup

Endpoint provider abstraction to decouple AWS/gcloud-specific logic from core daemon/proxy packages: `docs/plans/2026-04-14-endpoint-provider-abstraction-design.md`

## Test plan

- [x] Unit tests for credential provider (caching, refresh, context cancellation)
- [x] Unit tests for metadata endpoint handler (all routes, error classification)
- [x] Unit tests for grant config parsing and credential reconstruction
- [x] Unit tests for proxy routing (hostname, IP, no-handler fallback)
- [x] Unit tests for direct metadata routing (ping, path, flavor header guard)
- [x] Unit tests for normalizeSAPath (email normalization, passthrough, recursive)
- [x] Unit tests for FindGCloudHandler (single run, same profile, different profiles, mixed, no gcloud)
- [x] Persistence round-trip test with GCloudConfig
- [x] `go build ./...` passes
- [x] `make test-unit` passes (pre-existing env-dependent failures in container/deps only)
- [x] `make lint` clean for all changed files
- [x] E2E: `moat grant gcloud --project <p>` + `moat run --grant gcloud` + `gcloud projects list` inside container


dpup:
Addressed all four items from the latest review in aee6aaf:

1. **Relative key file path** — `filepath.Abs` before storing in credential metadata. Both `handleRegisterRun` and `RestoreRuns` paths now get absolute paths.
2. **Zero-expiry token cache** — Added `p.cached.Expiry.IsZero()` guard so tokens without expiry are treated as indefinitely valid.
3. **SA listing duplication** — Only list email-based account when `h.email != DefaultEmail`. Added test for the `DefaultEmail` listing path.
4. **`--scopes` in guide** — Added "Custom scopes" subsection with example.

Mutex-across-network-IO remains deferred as before (low priority for this use case).

Issue #320, opened 2026-04-15
tests: close coverage gaps for host-traffic proxy path


dpup:
## Summary

Code review of #NNN (fix/host-traffic-proxy-bypass branch) flagged two testing gaps that don't block the fix but should be closed for regression safety.

## Gaps

1. **P2-7**: \`proxy.go\` host-gateway aliasing has unit coverage for name/IP aliases, but no integration test asserts that a container setting \`MOAT_HOST_GATEWAY\` to a crafted bypass value (e.g. \`127.0.0.1\`, \`localhost\`, \`::1\`) is still forced through the proxy. Add an e2e case for this.

2. **P2-9**: \`addProxyPortForLoopback\` dedup path is covered for \`HostGateway == "127.0.0.1"\` but not for the \`moat-host\` branch (still relevant even after the recent revert — make sure the helper is parameterized so future extensions stay covered). Include a test where the proxy port is already present in \`AllowedHostPorts\` and assert no duplicate is added.

## Suggested owner

proxy / networking

## Impact

No functional gap today; tightens the net against regressions on the bypass fix landed in #NNN.

Issue #319, opened 2026-04-15
proxy: rewrite outbound URL host via url.Parse, not strings.Replace


dpup:
## Summary

In \`internal/proxy/proxy.go\`, three call sites rewrite the host portion of outbound URLs with \`strings.Replace(outURL, host, rc.HostGatewayIP, 1)\`. This is a substring match, not a structural URL edit — if the host substring appears elsewhere in the URL (path, query, userinfo), the rewrite will corrupt the request.

Not triggerable today with the \`moat-host\` synthetic hostname, but fragile and surprising. Would become a real bug the first time we rename the synthetic host or the gateway resolves to a value that overlaps path/query text.

## Context

Surfaced during code review of #NNN (fix/host-traffic-proxy-bypass branch) as P2-1 in the CE review. The fix is straightforward but deserves its own PR with targeted tests (path/query containing the host literal, percent-encoded variants).

## Call sites

- \`internal/proxy/proxy.go:1264\`
- \`internal/proxy/proxy.go:1407\`
- \`internal/proxy/proxy.go:1528\`

## Suggested fix

\`\`\`go
u, err := url.Parse(outURL)
if err != nil { /* preserve original on failure */ }
u.Host = rc.HostGatewayIP
outURL = u.String()
\`\`\`

Leave a TODO at the call sites if we land the refactor separately.

dpup:
This was already fixed in #321 (commit cb2bf14). All three call sites in `internal/proxy/proxy.go` now use the structural `rewriteURLHost` / `rewriteHostPort` helpers instead of `strings.Replace`. Unit tests covering IPv6 brackets, port preservation, path/query not being corrupted, and malformed URL fallback are in `TestRewriteURLHost` and `TestRewriteHostPort`.

Issue #318, opened 2026-04-15
proxy: firewall exec-attach race — fast containers exit before iptables rules apply


dpup:
## Summary

The egress firewall is applied via \`docker exec\` (or equivalent) after the container starts. Fast-exiting containers — e.g. \`moat run -- sh -c 'curl -s http://bad.example >/dev/null'\` — can complete their outbound request before the iptables rules are installed, giving a narrow window where traffic that should have been blocked is allowed.

## Context

Surfaced during code review of #NNN (fix/host-traffic-proxy-bypass branch) as P1-5 in the CE review. The branch adds a \`sleep 1\` to one e2e test as a workaround for this race; production code is unmitigated. Tracked for a separate root-cause fix.

## Repro

1. \`moat run\` with \`network.policy: strict\` and a command that makes a network request and exits immediately (e.g. \`curl\` + \`exit\`).
2. On a few runs, traffic to a non-allowed host succeeds because the firewall hadn't finished installing.

## Suggested fix

- Start the container paused, install iptables rules, then unpause.
- Or apply the firewall inside the entrypoint (before \`exec\` to the user command) so rules are guaranteed to be in place.
- Add a hostile test that exits in <50ms and asserts the block always takes effect.

## Impact

Narrow race window under strict policy. The fix should make strict policy unconditionally safe regardless of how fast the container exits.

Issue #317, opened 2026-04-15
proxy: IPv6 egress not firewalled — ip6tables rules not installed


dpup:
## Summary

The iptables-based egress firewall installed by the proxy covers IPv4 only. Containers with IPv6 connectivity can bypass `network.host` and `network.policy: strict` by using IPv6 addresses.

## Context

Surfaced during code review of #NNN (fix/host-traffic-proxy-bypass branch) as P0-2 in the CE review. Scoped out of that PR so we could ship the v4 bypass fix and handle IPv6 separately with proper test coverage.

## Repro

1. `moat run` with `network.policy: strict` in an environment where the container gets an IPv6 address (or where the resolver returns a AAAA record).
2. Run `curl -6 https://api.github.com/` (or any other blocked host) from inside the container.
3. Traffic succeeds — there are no `ip6tables` rules installed to match the `iptables` rules.

## Expected

IPv6 egress should be subject to the same deny-by-default posture and `network.host` / `network.rules` allowlist as IPv4.

## Suggested fix

- Mirror the iptables rule installer against `ip6tables`.
- Add an integration test that asserts `curl -6` to a blocked host fails under strict policy.
- Decide whether to emit a warning (or refuse to start) when `ip6tables` is unavailable in the host image.

## Impact

Silent bypass of network policy on dual-stack hosts. Severity P0 because the posture the user configured is not actually enforced.

PR #321, opened 2026-04-15
fix(proxy): prevent host traffic from bypassing proxy

dpup:
## Summary

- Host traffic no longer bypasses the proxy via `NO_PROXY=$MOAT_HOST_GATEWAY`. Container traffic to the host now flows through the proxy under the synthetic names `moat-host` (host services) and `moat-proxy` (proxy access), and `HTTP_PROXY` / `NO_PROXY` / `ALL_PROXY` / `MOAT_HOST_GATEWAY` / `MOAT_EXTRA_HOSTS` supplied via `moat.yaml env:` or `-e` are filtered with a warning so users cannot reopen the bypass.
- `moat-init.sh` fails closed on Apple containers when `/etc/hosts` cannot be written so strict policy does not silently degrade, and the image-tag hash includes the init-script SHA256 so cache invalidation is correct.
- Daemon advertises a `host-gateway-v2` capability; `moat run` fails fast with an actionable error against an older daemon (`run 'moat proxy restart'`). This prevents version skew from silently breaking host routing.
- Proxy registration now triggers for grant-less runs that configure `network.host`, `network.rules`, MCP servers, or Keep policies — previously these features were silently inert on permissive no-grant runs.
- CHANGELOG and `docs/content/reference/02-moat-yaml.md` call out the `MOAT_HOST_GATEWAY` IP→hostname semantic change so container scripts that passed the value to `ping` / `nslookup` / `iptables` can migrate.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `golangci-lint run ./...` clean
- [x] Unit tests pass in `internal/daemon`, `internal/proxy`, `internal/run`, `internal/deps`, `internal/hostnames` (only `TestRegistryGithubBinaryURLsExist` failures are pre-existing sandbox-network flakes)
- [x] New unit tests: `TestIsHostGatewayAliases`, `TestRedactURLUserinfo`, updated `TestAddProxyPortForLoopback_MoatHost` asserting the revert, extended env-filter test
- [ ] E2E: run `go test -tags=e2e ./internal/e2e/host_traffic_test.go` on a host with Docker
- [ ] Manual: `moat run -- curl http://$MOAT_HOST_GATEWAY:5432` against an allowed host port with `network.policy: strict` and no grants

## Follow-ups (filed as separate issues)

- #317 — IPv6 egress / ip6tables gap
- #318 — firewall exec-attach race
- #319 — URL rewrite via `strings.Replace`
- #320 — testing coverage gaps


dpup:
Thanks for the review. Addressed in cb2bf14:

1. **IPv6 URL rewrite bug** — real bug I introduced when I added `::1` to the `isHostGateway` alias set. Replaced the three `strings.Replace` call sites with `rewriteURLHost` (full URLs, for `handleHTTP`) and `rewriteHostPort` (dial addresses, for `handleConnectTunnel` / `handleConnectWithInterception`). Both use `net/url` + `net.JoinHostPort` so IPv6 brackets are handled correctly and path/query text that happens to match the host literal is not corrupted. Also closes the `strings.Replace` refactor tracked in #319 for the three call sites touched here. Covered by `TestRewriteURLHost` and `TestRewriteHostPort`.
2. **`host-gateway-v2` hard-fail** — intentional: every proxy run sets `HostGateway = moat-host` and `HostGatewayIP = 127.0.0.1`, so an older daemon would silently route host traffic wrong. The CHANGELOG entry already leads with "Run `moat proxy restart` after upgrading to pick up this fix"; happy to add more prominence if you want.
3. **`localhost`/`127.0.0.1` NO_PROXY bypass under Docker host networking** — good catch, confirmed pre-existing. Filed as majorcontext/moat#322 with three fix options (drop from NO_PROXY + rely on firewall, apply firewall unconditionally, or document the gap) so we can decide deliberately rather than rushing it into this PR.
4. **Comment typo** — dismissed: the comment at manager.go:1292 is a proper `//` comment. Looks like a rendering artifact.

dpup:
Addressed in ec9739a — bracket bare IPv6 in `rewriteURLHost` when the URL has no port, and added a `{"http://moat-host/x", "::1"} → "http://[::1]/x"` test case. Unreachable today but correct for when IPv6 gateways land via #317.

dpup:
Addressed in 3112bc8 — dropped the redundant MCP policy loop; the preceding `len(opts.Config.MCP) > 0` already covers it. E2E failure on the prior run was a CI infrastructure flake ("The hosted runner lost communication with the server" after 1h47m), not a test failure — the push should re-trigger it.

PR #323, opened 2026-04-16
feat(config): honor MOAT_HOME to relocate moat state directory

dpup:
## Summary

Introduces `MOAT_HOME` as the single override for the moat configuration directory (default `~/.moat`). When set, it replaces `~/.moat` as the root for runs, credentials, daemon socket/lock, keyring key file, per-run SSH sockets, and the routing proxy state.

Third-party state (`~/.claude`, `~/.config/gh`, etc.) still resolves against the real `$HOME` — this is only about **moat-owned** paths.

## Motivation

Two narrow use cases, both rare:

1. **Hermetic e2e tests.** Without isolation, tests share the developer's daemon, credential store, run storage, and lock files with any live moat session. An older daemon from a previous install can serve test requests and fail on missing capabilities; test cleanup (`killTestDaemon`) can kill the user's working daemon. The e2e `TestMain` now creates a temp `MOAT_HOME` so all daemon/proxy/credential state lands in an isolated directory that's deleted on exit.
2. **Running two versions of moat side-by-side**, each with its own state.

## Design

- Single entry point: `config.GlobalConfigDir()` honors `MOAT_HOME` if set, otherwise falls back to `$HOME/.moat`.
- Every moat-owned path lookup routes through it: `storage.DefaultBaseDir`, `credential.DefaultStoreDir`, credential keyring paths, oauth config, deps/versions cache, configprovider loader, `run.manager` SSH socket dir, and CLI commands (audit, doctor, volumes).
- The daemon subprocess inherits `os.Environ()`, so `MOAT_HOME` propagates across the fork.
- Keyring preserves its no-tempdir-fallback security invariant — `MOAT_HOME` is layered as an early return above the existing error chain.
- Tests that pin `HOME` to a tempdir now also clear `MOAT_HOME` (via `t.Setenv`) so a leaked override from the parent environment cannot redirect assertions.

## e2e port isolation

Second commit: the routing proxy defaults to port 8080, a global listen port that is **not** covered by `MOAT_HOME` state isolation. When a developer has a live moat routing proxy on 8080, the e2e test harness's fresh proxy cannot bind it. Setting `MOAT_PROXY_PORT=0` in `TestMain` tells the routing proxy to let the OS pick a free port. Tests already read the actual port from the routing proxy lock file, so whichever port lands is transparent. The daemon's credential proxy (19080 default) already has fallback-to-OS-port logic when its port is taken.

## Test plan

- [x] Unit tests pass on every touched package (config, storage, credential, credential/keyring, providers/oauth, providers/configprovider, deps/versions, run, cmd/moat/cli)
- [x] Unit tests pass with `MOAT_HOME=/tmp/leaked` leaked into the parent environment (no cross-test leakage)
- [x] `make lint` — 0 issues
- [x] Previously-failing e2e endpoint routing tests now pass: `TestEndpointRouteRegistration`, `TestEndpointRoutingBasic`, `TestEndpointRoutingMultiEndpoint`, `TestEndpointRoutingWithTLS`, `TestEndpointRoutingWithGrants`
- [x] New unit tests verify `MOAT_HOME` override for `storage.DefaultBaseDir` and `keyring.DefaultKeyFilePath`
- [x] Docs updated (`docs/content/reference/03-environment.md` — new `### MOAT_HOME` section)

### DIFF
--- a/cmd/moat/cli/audit.go
+++ b/cmd/moat/cli/audit.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/majorcontext/moat/internal/audit"
 	"github.com/majorcontext/moat/internal/run"
+	"github.com/majorcontext/moat/internal/storage"
 	"github.com/majorcontext/moat/internal/ui"
 	"github.com/spf13/cobra"
 )
@@ -64,11 +65,7 @@ func runAudit(cmd *cobra.Command, args []string) error {
 	}
 
 	// Find run directory
-	homeDir, err := os.UserHomeDir()
-	if err != nil {
-		return fmt.Errorf("finding home directory: %w", err)
-	}
-	runDir := filepath.Join(homeDir, ".moat", "runs", runID)
+	runDir := filepath.Join(storage.DefaultBaseDir(), runID)
 	dbPath := filepath.Join(runDir, "audit.db")
 
 	if _, statErr := os.Stat(dbPath); os.IsNotExist(statErr) {
--- a/cmd/moat/cli/doctor.go
+++ b/cmd/moat/cli/doctor.go
@@ -19,6 +19,7 @@ import (
 	"github.com/majorcontext/moat/internal/credential"
 	"github.com/majorcontext/moat/internal/doctor"
 	"github.com/majorcontext/moat/internal/providers/codex"
+	"github.com/majorcontext/moat/internal/storage"
 	"github.com/majorcontext/moat/internal/ui"
 	"github.com/spf13/cobra"
 )
@@ -410,7 +411,7 @@ func (s *claudeSection) Print(w io.Writer) error {
 	}
 
 	// 3. Moat-specific user defaults
-	moatUserSettingsPath := filepath.Join(home, ".moat", "claude", "settings.json")
+	moatUserSettingsPath := filepath.Join(config.GlobalConfigDir(), "claude", "settings.json")
 	if _, err := os.Stat(moatUserSettingsPath); err == nil {
 		fmt.Fprintf(tw, "  3. Moat user defaults:\t%s %s\n", moatUserSettingsPath, ui.OKTag())
 	} else {
@@ -452,13 +453,8 @@ type storageSection struct{}
 func (s *storageSection) Name() string { return "Storage" }
 
 func (s *storageSection) Print(w io.Writer) error {
-	home, err := os.UserHomeDir()
-	if err != nil {
-		return nil
-	}
-
 	tw := tabwriter.NewWriter(w, 0, 0, 2, ' ', 0)
-	moatDir := filepath.Join(home, ".moat")
+	moatDir := config.GlobalConfigDir()
 	fmt.Fprintf(tw, "Moat directory:\t%s\n", moatDir)
 
 	if info, err := os.Stat(moatDir); err == nil {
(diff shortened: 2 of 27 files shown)


dpup:
Thanks for the review. Addressed in f5aca71:

- **Bug (keyring.go:327)** — error message now uses `filepath.Dir(fallback.Name())` so the remediation path matches wherever the key was being written.
- **Missed callsite #1 (claude/settings.go:418)** — `moatUserSettingsPath` now uses `config.GlobalConfigDir()`. `moat doctor` and `moat run` will now agree.
- **Missed callsite #2 (run/services.go:224)** — service dependency cache path now uses `config.GlobalConfigDir()`. Dropped the `os.UserHomeDir()` call entirely in that block.

Deferred (with rationale):

- **Keyring MOAT_HOME duplication** — you noted this is acceptable as follow-up. The consolidation is non-trivial because keyring's HOME-fallback has a stricter security posture than `config.GlobalConfigDir()` (refuses `./.moat` fallback and errors instead). Cleanest path is a separate `config.SecureGlobalConfigDir() (string, error)` helper, which I'd rather do in a dedicated PR.
- **DefaultStoreDir fallback behavior change** — agreed the new behavior is safer. I'd rather not add a "why we removed the tempdir fallback" comment here since the old behavior is gone and the keyring doc already covers the temp-dir-is-unsafe rationale.
- **MOAT_HOME absolute-path validation** — low-priority per your note. Happy to add this if you'd rather have it; otherwise leaving it for future hardening.

dpup:
Thanks, second round. All three fixed in f250094:

- **settings_test.go missing MOAT_HOME clear** — added `t.Setenv("MOAT_HOME", "")` to both `TestLoadAllSettingsSkipHostSettings` and `TestLoadAllSettingsPreservesMoatUserExtras`. Verified the latter passes with `MOAT_HOME=/tmp/leaked` in the parent env.
- **`run/manager.go:589`** — warning now uses `filepath.Join(config.GlobalConfigDir(), "config.yaml")`.
- **`aws/endpoint.go:207,237`** — computed `daemonLog` once at the top of `classifyAWSError` and substituted into both messages. Imported moat's config as `moatconfig` to avoid colliding with aws-sdk-go-v2's `config` package already imported here.
- **Minor keyring package doc** — added "or `\$MOAT_HOME/...` when MOAT_HOME is set" qualifiers to lines 7 and 14.

dpup:
Round 3 fixes pushed in c8b9eb9:

- `cmd/moat/cli/grant_providers.go` — custom providers path now routes through `config.GlobalConfigDir()`
- `cmd/moat/cli/grant_oauth.go` — "no OAuth config found" error now formats the real config path via `oauth.DefaultConfigDir()`
- `internal/worktree/repo.go` — `BasePath` falls back to `config.GlobalConfigDir()/worktrees` when `MOAT_WORKTREE_BASE` is unset
- `internal/providers/claude/settings.go` — `SourceMoatUser` const changed from a hardcoded path string to the stable label `"moat user settings"` (verified no JSON/persistence dependency)

`go build`, `make lint`, and tests on affected packages all green.

dpup:
Round 4 fixes pushed in d5f4d35 — both minor findings addressed:

- **`keyring.go` globalLockPath** — now mirrors `DefaultKeyFilePath`'s MOAT_HOME-aware resolution and returns `ErrNoHomeDirectory` instead of falling back to `os.TempDir()`. Signature changed to `(string, error)`; only caller (`withGlobalKeyLock`) propagates the error. Fixed the asymmetry you called out rather than deferring it.
- **`configprovider/loader.go:71`** — stale doc comment updated to `<MOAT_HOME>/providers/ (default ~/.moat/providers/)`.

`make lint` and keyring + configprovider tests all green.

dpup:
Good catch — real bug. Fixed in 9f01c03:

`fileBackend.Name()` returns a display string `"file (/path/to/key)"`, so my round-1 `filepath.Dir(fallback.Name())` produced garbled output. Took the type-assertion approach you suggested — falls back to a generic `"moat key directory"` label if the backend isn't a `*fileBackend` so the error message stays well-formed regardless.

Lint + keyring tests green.

PR #353, opened 2026-05-19
fix(claude): don't scrape claude setup-token output

natehardison:
moat grant claude spawned `claude setup-token` under a PTY, captured its output, and parsed the OAuth token out of it. Recent Claude CLI versions render setup-token as a TUI, so the token never appeared contiguously in the captured stream and extraction always failed with "could not find OAuth token in claude setup-token output", forcing a manual paste anyway.

Any output scraping here is inherently brittle: the CLI's rendering is not a stable interface and changes between versions. Instead, run `claude setup-token` attached directly to the user's terminal (no PTY, no capture) so Claude displays the token as normal, then have the user paste it and validate it against the API.

A token soft-wrapped by a narrow terminal puts real newlines in the clipboard. On an interactive terminal the paste is read via bracketed paste (DECSET 2004): the whole paste is captured between ESC[200~ and ESC[201~ with no terminator keystroke, then whitespace is stripped (tokens contain none). Non-interactive/piped input falls back to reading until a blank line or EOF. The scanner is a pure function over an io.Reader and is unit-tested without a terminal.

Consolidates the paste-and-validate logic shared with the existing-token option into one helper.

### DIFF
--- a/internal/providers/claude/grant.go
+++ b/internal/providers/claude/grant.go
@@ -17,6 +17,7 @@ import (
 	"github.com/majorcontext/moat/internal/log"
 	"github.com/majorcontext/moat/internal/provider"
 	"github.com/majorcontext/moat/internal/term"
+	"github.com/majorcontext/moat/internal/ui"
 )
 
 // Grant acquires a Claude Code OAuth token interactively.
@@ -74,10 +75,10 @@ func (p *OAuthProvider) Grant(ctx context.Context) (*provider.Credential, error)
 				fmt.Printf("Invalid choice: %s\n", response)
 				continue
 			}
-			return grantViaSetupToken(ctx)
+			return grantViaSetupToken(ctx, reader)
 
 		case fmt.Sprint(existingTokenOpt):
-			return grantViaExistingOAuthToken(ctx)
+			return grantViaExistingOAuthToken(ctx, reader)
 
 		case fmt.Sprint(importCredsOpt):
 			if importCredsOpt == 0 {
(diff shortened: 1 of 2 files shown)


natehardison:
I initially had Claude do a full extraction using VT emulation, so if you'd prefer that approach, the commit is here: https://github.com/majorcontext/moat/compare/main...natehardison:moat:fix/claude-setup-token-vt-extraction. It seemed to me to be more error-prone over the long term than this approach.

dpup:
Thanks for the patch. Let me look into this next week.

dpup:
Pushed two maintainer commits and rebased onto `main`.

**Fix — piped paste dropped the token.** `Grant()` reads the menu choice with a `bufio.Reader` over `os.Stdin`; on piped input that reader buffers the token line too, and `promptAndValidateOAuthToken` then created a *second* `bufio.Reader` on `os.Stdin`, reading an already-drained stream — so a piped token came back empty (`OAuth token cannot be empty`). Repro: `printf '2\n<token>\n\n' | moat grant claude`. Threaded the menu's reader through to the paste read (extracted `readOAuthTokenInput`); the interactive TTY path is unchanged. Added a regression test (`TestReadOAuthTokenInput_pipedUsesSharedReader`).

**UX — separated moat's prompt from the `setup-token` output.** A `ui.Section` header now precedes the paste prompt so Claude's "…`export CLAUDE_CODE_OAUTH_TOKEN=`…" output doesn't run straight into moat's prompt; also dropped the now-redundant prompt line in the paste-existing-token flow.

The bracketed-paste approach itself is unchanged. Verified both paths at runtime: interactive (PTY, soft-wrapped token reassembled) and piped. The rebase onto `main` also preserves #358's one-line `subscriptionType` threading in `grantViaExistingCreds`.

PR #354, opened 2026-05-19
feat(kiro): add Kiro CLI agent support

natehardison:
## Summary

  Adds the Kiro CLI as a first-class Moat agent, at parity with the existing
  Codex/Claude/Gemini providers: credential grant, transparent proxy credential
  injection, container config staging, local + remote MCP, runtime-context
  injection, and a `moat kiro` command.

  ## What's included

  - **`internal/providers/kiro/`** — agent + credential provider mirroring the
    codex provider: proxy Bearer injection, token grant (env `KIRO_API_KEY` or
    interactive hidden-input prompt), `PrepareContainer` staging
    `~/.kiro/{settings,agents,steering}`, and the `moat kiro` command.
  - **Wiring** — `kiro-cli` dependency (registry + native installer),
    `credential.ProviderKiro`, `kiro:` moat.yaml section (`sync_logs`, `mcp`) +
    `ShouldSyncKiroLogs`, image-needs detection, `manager.go` dispatch block
    (local + remote MCP relay, grant→env maps, cleanup/temp-dir lifecycle),
    provider registration, and the `moat-init.sh` block that copies the staged
    `~/.kiro` tree into the container.
  - **Docs** — CLI reference, moat.yaml reference, a new guide
    (`docs/content/guides/14-kiro.md`), and a CHANGELOG entry. Design spec and
    implementation plan under `docs/plans/`.

  ## Notable design decisions

  - **Auth model:** prompt-for-token (or `KIRO_API_KEY` env), stored encrypted;
    static credential, no refresh — re-grant on expiry. The container only ever
    sees a placeholder; the proxy injects the real token.
  - **Credential-injection host scoping:** gatekeeper v0.2.0 credential
    injection uses exact host-key lookups (no wildcard matching — verified
    against `proxy.getCredentials`). `kiroAPIHosts` is therefore scoped to the
    concrete `q.us-east-1.amazonaws.com`; additional regions are added
    explicitly (documented in code and the guide).
  - **Remote MCP** uses kiro-cli's native HTTP MCP server entry
    (`url` + `headers`).

  ## Out of scope (intentional, documented in the spec)

  Token refresh / OAuth device-login, persistent sessions volume, host
  `~/.kiro/skills` layering, kiro subagents in the default agent, and the
  `moat init` quickstart table (separate hand-maintained list — tracked as a
  follow-up).

  ## Testing

  - `go build ./...` clean; `make lint` 0 issues; all kiro-relevant packages
    pass under `-race`.
  - The only failing unit test is the pre-existing
    `TestRegistryGithubBinaryURLsExist`, which makes live requests to
    github.com release URLs and fails in network-restricted environments
    (fails identically on `main`; unrelated to this change).
  - Manually validated `moat kiro` end-to-end.

  ## Test plan

  - [x] `moat grant kiro` stores a token
  - [x] `moat kiro` runs an isolated session with credentials injected
  - [ ] `kiro.mcp` (local) and top-level `mcp:` (remote relay) reach the agent
  - [x] runtime context appears via `~/.kiro/steering/`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Moat is pre-1.0. The CLI interface and `moat.yaml` schema may change between min
 
 ## Unreleased
 
+### Added
+
+- **Kiro CLI support** — `moat kiro` runs the Kiro CLI in an isolated container with transparent credential injection; `moat grant kiro` stores a Kiro token. Supports local and remote MCP and runtime-context injection. ([#NNN](https://github.com/majorcontext/moat/pull/NNN))
+
 ## v0.5.1 — 2026-04-28
 
 Patch release with one security fix (IPv6 egress firewall) and a batch of run-lifecycle and proxy fixes. Adds `MOAT_HOME` for relocating moat state, a multi-runtime manager so Docker and Apple containers can coexist in one install, TUI debug shortcuts, and Python 3.13/3.14 support. Gatekeeper is extracted to its own repository.
--- a/docs/content/guides/14-kiro.md
+++ b/docs/content/guides/14-kiro.md
@@ -1,0 +1,190 @@
+---
+title: "Running Kiro"
+navTitle: "Kiro"
+description: "Run the Kiro CLI in an isolated container with credential injection."
+keywords: ["moat", "kiro", "ai agent", "coding assistant"]
+---
+
+# Running Kiro
+
+This guide covers running the Kiro CLI in a Moat container.
+
+## Prerequisites
+
+- Moat installed
+- A Kiro API token
+
+## Granting Kiro credentials
+
+Run `moat grant kiro` to configure authentication:
+
+```bash
+$ moat grant kiro
+
+Enter your Kiro token: ...
+
+Kiro token saved to ~/.moat/credentials/kiro.enc
+```
+
+You can also set `KIRO_API_KEY` in your environment before running the command:
+
+```bash
+export KIRO_API_KEY="..."
+moat grant kiro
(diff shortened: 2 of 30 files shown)


dpup:
Thanks for the patch. A few things before merging:

**Memory bug:** `isAIAgent()` in manager.go doesn't include kiro, so on Apple containers it skips the 8GB agent default and drops to 4GB. Kiro's exactly the kind of workload that'll OOM on a big task — quick fix.

**Auth path:** we're only injecting on `q.us-east-1.amazonaws.com` with a static Bearer. Did you confirm that's actually where kiro-cli authenticates, and that a plain Bearer works there (vs SigV4)? And if there's any token handshake at cli.kiro.dev/cognito, those aren't getting injection. I don't want to quietly 401 for anyone outside the path you tested — other regions especially. Also, wherever these host/auth assumptions are called out in the comments, can you link the kiro docs you're basing them on so it's checkable later?

Same spirit on the other kiro-cli assumptions the comments flag — the mcp.json key names and the `--no-interactive` prompt arg. Did your manual run cover a remote MCP round-trip and a `-p` prompt? If not, those are the bits I'd poke at.

Couple small ones: the CHANGELOG still has the `#NNN` placeholder, and it'll need a rebase onto main — a fair bit landed since you branched (grant.go and manager.go both moved), so worth doing carefully.

Let's also fold these into this PR rather than follow-ups, since we're calling kiro a first-class agent: it shows up blank in `moat grant providers`, and `init`/`doctor` don't know about it yet.

PR #362, opened 2026-05-28
fix(tui): drain vt emulator reply pipe to prevent compositor deadlock

abezzub-dr:
Fixes moat hanging on startup every session. Not sure what changed, the issue started today and I can't use moat without this fix. Description by Claude below.

## Summary

When the child inside the container emits any terminal query that requires a reply (CSI c Primary Device Attributes, CSI 6n cursor position, in-band resize, color queries, etc.), moat's compositor freezes on the first paint. From the user's perspective the screen rendered the initial banner and then never updated again. Input still reached the child — so Ctrl+C typed twice into a "frozen" Claude actually terminated the container — but no further output reached the host terminal. With Claude Code 2.1.150+ this reproduces 100% of the time.

### Root cause

`vt.NewEmulator` constructs an internal `io.Pipe` for replies the emulator generates on the child's behalf. The handlers write replies into the pipe writer, expecting a consumer to drain the reader side. Moat never drained it. The first reply-bearing handler blocked inside `io.PipeWriter.Write` — reached from `tui.Writer.Write` under `w.mu` — so `renderLoop` could never reacquire the mutex to paint another frame.

Goroutine dump from a stuck session showed the exact chain:

```
docker.StartAttached.func1                ← io.Copy(opts.Stdout, resp.Reader)
  io.multiWriter.Write → trace.RecordingWriter.Write
    tui.Writer.Write                       (writer.go:171, HOLDS w.mu)
      processDataLocked → outputLocked
        vt.Emulator.Write → ansi.Parser.advance → handleCsi 'c'
          io.PipeWriter.Write              ← BLOCKED FOREVER (no reader)
```

And, in another goroutine, `renderLoop [sync.Mutex.Lock, 3 minutes]`.

### `internal/tui/writer.go`

- Add `Injector` interface (single method `Inject([]byte) error`) and an `injector` field on `Writer`, plus `SetInjector` to wire it up.
- On compositor entry, spawn `drainEmulatorReplies` goroutine that copies `em.Read` → `injector.Inject`. Bytes are still drained if no injector is wired, so the deadlock can't recur regardless of caller plumbing — the child just doesn't see its replies in that case and falls back to defaults.
- On compositor exit / `Reset` / `Cleanup`: close the emulator's input pipe via `em.InputPipe().(io.Closer).Close()` to unblock `Read` and let the drain goroutine exit. We deliberately avoid `vt.Emulator.Close()` — it writes `e.closed` unsynchronized with concurrent `Read` calls and trips `-race`. The pipe close is documented as safe for parallel Close-vs-Read and achieves the same effective shutdown.

### `cmd/moat/cli/exec.go`

- After constructing the `InjectableReader` for the stdin chain, call `statusWriter.SetInjector(injectable)`. The injectable already exists for synthetic keystrokes (Ctrl+L from `resetTUI`), so emulator replies travel the same recorded path and appear in trace dumps alongside user input.

## Test plan

- [x] New test `TestWriter_CompositorMode_DeviceAttributesQuery_DoesNotDeadlock` — sends `CSI c` in compositor mode, fails (deadlock) without the fix, passes with it.
- [x] New test `TestWriter_CompositorMode_InjectsEmulatorReplies` — verifies the Primary DA reply (begins with `CSI ?`) reaches the configured injector.
- [x] `go test -race ./internal/tui/` clean.
- [x] `make test-unit` passes.
- [x] Manual: rebuilt binary, ran `moat claude` in a project with a `moat.yaml` that triggered the freeze 100% of the time before this PR — Claude now starts and accepts input normally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Moat is pre-1.0. The CLI interface and `moat.yaml` schema may change between min
 
 ## Unreleased
 
+### Fixed
+
+- Fix Claude Code (and other Ink-based TUIs) freezing on the first paint inside the moat container — previously, when the child emitted any terminal-capability query that required a reply (CSI c Primary Device Attributes, CSI 6n cursor position, in-band resize, color queries), moat's VT compositor blocked forever on the emulator's internal reply pipe with no drain, holding `tui.Writer.mu` and starving the render goroutine. Input still reached the child, but no further output reached the host terminal, so the session appeared frozen and Ctrl+C wasn't visibly acknowledged even though it terminated the child. Surfaced 100% of the time with Claude Code 2.1.150+, which queries Device Attributes at startup. Moat now drains the emulator's reply pipe and routes replies back into the child's input stream via the existing injectable-reader chain. ([#362](https://github.com/majorcontext/moat/pull/362))
+
 ## v0.5.3 — 2026-05-25
 
 Patch release centered on Claude Code authentication inside containers — subscription detection, `setup-token` capture, and version pinning — plus a non-root tmpfs permissions fix.
--- a/cmd/moat/cli/exec.go
+++ b/cmd/moat/cli/exec.go
@@ -445,6 +445,14 @@ func RunInteractiveAttached(ctx context.Context, manager *run.Manager, r *run.Ru
 	injectable := term.NewInjectableReader(stdin)
 	defer injectable.Close()
 	stdin = injectable
+
+	// Wire the injectable to receive VT-emulator reply bytes (Primary DA,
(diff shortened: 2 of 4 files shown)


dpup:
Nice find. Have you been able to manual test in both docker and apple containers?

abezzub-dr:
Claude starts with both docker and apple containers.

vanakema:
@dpup would it be possible to get a release cut with this fix ASAP? Sucks to not be able to use the new version of claude w/ moat rn.

dpup:
@vanakema yeah, let me do that now. Sorry. The weird thing is it stopped hitting me...

vanakema:
I see the release was cut; thank you so much @dpup!

Issue #375, opened 2026-06-16
run stdout not consistently streamed to the terminal inline (only reliable via `moat logs`)


dpup:
## Summary

`moat run` sometimes prints the container's stdout inline after `Press Ctrl+C to stop`, and sometimes prints nothing — the output is then only retrievable via `moat logs <run-id>`. Same command, same example, different runs.

## Repro

- `moat run examples/grant-github-nossh` repeatedly. Some invocations show the script output inline; others show only the `Started …` / `View output:` lines and require `moat logs <id>` to see anything.

## Environment

- macOS, ghostty terminal (cmux).

## Notes

- Observed across many non-interactive (no `-i`), short-lived container runs while validating #370. Possibly a race between container exit / log flush and the host attach/stream path (inline stream vs `storage.LogWriter` capture).

Issue #374, opened 2026-06-16
github: stale/invalid stored token fails with opaque `could not read Username` — surface a re-grant hint


dpup:
## Summary

When the stored GitHub token is invalid/expired, HTTPS git fails with:

```
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

and the proxy logs a bare `401 invalid credentials`. Nothing indicates the **granted credential** is the problem or that re-granting fixes it.

## Repro

- Have a `github` grant whose stored token has been rotated/revoked.
- `moat run --grant github -- git ls-remote https://github.com/<any>.git` → `could not read Username` (401 at the proxy).

## Expected

A 401 from `api.github.com` / `github.com` on an injected credential should produce an actionable hint, e.g.:

> GitHub rejected the injected credential (401). The stored token may be expired — run `moat grant github` to refresh.

Optionally validate the token at run start (the provider already validates at grant time).

## Notes

- Cost a long debugging detour while validating #370 — the error read like a code bug when it was a stale stored credential. Compounded by the daemon not reloading on re-grant (separate issue).

Issue #373, opened 2026-06-16
daemon serves a stale credential after re-grant until `moat proxy stop`


dpup:
## Summary

After `moat grant <provider>` updates a stored credential, an already-running proxy daemon keeps injecting the **previous** token. The new credential only takes effect after `moat proxy stop` (which forces a fresh daemon to reload from the credential store on the next run).

## Repro (observed with `github`)

1. `moat grant github` with a token that later becomes invalid (e.g. a rotated `gho_` OAuth token).
2. `moat grant github` again to store a fresh, valid token.
3. `moat run --grant github -- git ls-remote https://github.com/octocat/Hello-World.git HEAD` → still fails `401 invalid credentials` (old token is being injected).
4. `moat proxy stop`, then re-run the same command → succeeds.

## Expected

Re-granting a credential should propagate to the running daemon — either `moat grant` signals the daemon to reload, or the daemon re-resolves the credential from the store on each new run registration instead of serving a cached value.

## Notes

- Surfaced while validating #370. The daemon outlives the CLI by design, so a CLI re-grant never reaches it.
- Workaround today: `moat proxy stop` after re-granting.

PR #376, opened 2026-06-16
fix(github): use Basic auth for github.com git smart-HTTP over HTTPS

### LINKED ISSUE #370: github grant injects Bearer for github.com, but git smart-HTTP requires Basic — HTTPS git fetch/push still broken (follow-up to #180)
## Summary

#180 documented that git HTTPS operations fail through the proxy and was closed by routing git over SSH when an `ssh:github.com` grant is active. Without an SSH grant, the default HTTPS path is still broken, and the root cause is identifiable: the `github` provider injects the wrong auth scheme for `github.com`.

Tested on v0.5.4 (`moat claude --no-sandbox --grant github`, network policy permissive).

## Root cause (two stacked failures)

**1. git aborts the proxy CONNECT on the 407 challenge.**

curl sends `Proxy-Authorization` preemptively from the `https_proxy` URL credentials; git does not, and it doesn't retry after the 407:

```
=> Send header: CONNECT github.com:443 HTTP/1.1
=> Send header: Proxy-Connection: Keep-Alive
<= Recv header: HTTP/1.1 407 Proxy Authentication Required
fatal: unable to access 'https://github.com/owner/repo.git/': CONNECT tunnel failed, response 407
```

Remedy: `git config http.proxyAuthMethod basic` (as #180 also noted). Moat could set this in the container's gitconfig during init.

**2. GitHub's git smart-HTTP endpoints reject the injected Bearer header.**

`internal/providers/github/provider.go` injects `Authorization: Bearer <token>` for both `api.github.com` and `github.com`. The git endpoints (`info/refs`, `git-upload-pack`, `git-receive-pack`) return 401 for Bearer; they require Basic auth. Verified outside moat against a private repo:

```
$ git -c http.extraHeader="Authorization: Bearer $TOKEN" ls-remote https://github.com/owner/private-repo.git HEAD
fatal: could not read Username for 'https://github.com': ...   # server returned 401

$ B=$(printf "x-access-token:%s" "$TOKEN" | base64 -w0)
$ git -c http.extraHeader="Authorization: Basic $B" ls-remote https://github.com/owner/private-repo.git HEAD
da6a0ea2...	HEAD
```

And inside the sandbox (with fix 1 applied and a custom provider injecting Basic — see workaround): `curl` on `info/refs?service=git-upload-pack` goes 401 → 200, and `git ls-remote`/clone/push work.

## Suggested fix

In the github provider, inject the Basic form for the `github.com` host (this is what GitHub Actions' checkout does):

```go
basic := base64.StdEncoding.EncodeToString([]byte("x-access-token:" + cred.Token))
proxy.SetCredentialWithGrant("github.com", "Authorization", "Basic "+basic, "github")
// api.github.com stays Bearer
```

plus `http.proxyAuthMethod = basic` in the container gitconfig so git survives the CONNECT challenge. That would make HTTPS git work out of the box with just `--grant github`, no SSH key needed.

## Workaround we're using today

A custom provider whose stored "token" is the pre-encoded Basic blob (provider YAML can't base64-encode):

```yaml
name: github-git
hosts: ["github.com"]
inject:
  header: "Authorization"
  prefix: "Basic "
source_env: [GITHUB_GIT_BASIC]
```

```bash
GITHUB_GIT_BASIC=$(printf "x-access-token:%s" "$(gh auth token)" | base64 -w0) moat grant github-git
```

granted **before** `github` (the first grant appears to win the `github.com` host collision — if that ordering is intentional, documenting it would help), plus `git config http.proxyAuthMethod basic` in the workspace.

Happy to provide more traces if useful.

dpup:
## What

HTTPS `git` fetch/push to `github.com` failed with only the `github` grant (no SSH). Two stacked causes:

1. The provider injected `Authorization: Bearer <token>` for `github.com`, but GitHub's git smart-HTTP endpoints (`info/refs`, `git-upload-pack`, `git-receive-pack`) reject Bearer with **401** and require **Basic** `x-access-token:<token>` — the scheme GitHub Actions' checkout uses.
2. `git` aborts on the proxy's **407** CONNECT challenge because, unlike `curl`, it doesn't send `Proxy-Authorization` preemptively and doesn't retry.

## How

- **`internal/providers/github/provider.go`** — inject `Basic base64("x-access-token:<token>")` for `github.com` (Bearer stays for `api.github.com`), extracted into a shared `setProxyAuth()` helper.
- **`internal/providers/github/refresh.go`** — token refresh now reuses `setProxyAuth()` (it independently re-injected Bearer for `github.com`, which would have silently reverted the fix on the next refresh).
- **`internal/deps/scripts/moat-init.sh`** — set `git http.proxyAuthMethod=basic` so git clears the proxy 407.

HTTPS `git clone`/`fetch`/`push` now work with just `--grant github`. When both `github` and `ssh:github.com` are granted, git still routes over SSH as before — now framed as an identity preference (use the SSH key's identity), not a workaround. Stale comments and docs updated to match.

## Backwards compatibility

No change to `internal/daemon/api.go`. `CredentialSpec.Value` carries a `Basic …` string instead of `Bearer …` in the same existing field, so old/new daemon and CLI combinations stay compatible (the daemon stores and injects the value verbatim).

## Tests

- `TestProvider_ConfigureProxy` and both `Refresh` tests assert the per-host schemes (Bearer for API, Basic for git smart-HTTP).
- New `TestMoatInitScriptGitProxyAuth` asserts the init script sets `proxyAuthMethod=basic` and keeps the SSH `url.insteadOf` rewrite.
- `make lint` clean; full `-race` unit suite passes.

## Verification

Validated end-to-end on a real host (macOS) with a real `gho_` token via the new `examples/grant-github-nossh` example:

```
http.proxyAuthMethod = basic          PASS
HTTPS ls-remote ... HEAD = 7fd1a60b... PASS
HTTPS shallow clone ...                PASS
RESULT: PASS -- HTTPS git works with only --grant github
```

`api.github.com` (Bearer) and `github.com` git (Basic) both authenticate with the same grant. Negative control on `main` fails with `407 CONNECT tunnel failed`.

## Docs & example

- `docs/content/guides/04-ssh.md`, `docs/content/reference/04-grants.md` updated.
- `examples/grant-github-nossh/` added — runnable, self-checking verification with an optional `VERIFY_REPO` write test.

## Follow-ups (separate issues, surfaced during validation)

- #373 — daemon serves a stale credential after re-grant until `moat proxy stop`
- #374 — github: stale/invalid stored token fails with opaque `could not read Username`
- #375 — run stdout not consistently streamed inline

Closes #370


dpup:
Thanks — good catch on the indirect deployment chain. Verified it's actually already guarded at both critical links: `TestProvider_ImpliedDependencies` asserts `github → [gh, git]`, and `TestImageSpecNeedsInit`'s `GitIdentity` case asserts `NeedsGitIdentity → needsInit=true` (so moat-init.sh deploys). So rather than add a redundant test, I added a comment at `NeedsGitIdentity: hasGit` (manager.go) making the `github grant → git dep → init script → proxyAuthMethod` chain explicit and pointing at those two tests, so a refactor can't break it silently. Pushed in 4f947f9.

The `SourceCLI` Refresh gap you noted is tracked as a known low-priority item (the shared `setProxyAuth` helper means the injection logic itself is covered by the env-source and ConfigureProxy tests; only the `gh auth token` exec plumbing is untested).

PR #377, opened 2026-06-16
fix(container): report pre_run hook failures clearly

### LINKED ISSUE #372: Hook failures are hard to debug
When a hook fails, especially the pre_run hook, the error is hard to see and it looks like the container just failed to start.

dpup:
## What

Closes #372. A failing `hooks.pre_run` command looked like the container itself failed to start, making the real error hard to find.

`moat-init.sh` runs with `set -e`, and `run_pre_run_hook()` ran the hook with no exit-code check — so a non-zero hook exit aborted the entrypoint with a bare status *before* `exec`-ing the main command, with no framing to say the pre_run hook was the cause.

## Before / after

Reproduced with a `moat.yaml` whose `hooks.pre_run` runs a missing command:

**Before**
```
sh: 1: nonexistent-build-tool: not found
Error: run failed: exit code 127
```

**After**
```
moat: pre_run hook failed (exit code 127)
moat:   command: echo 'doing project setup...'; nonexistent-build-tool --install
moat:   the pre_run hook runs as moatuser in /workspace before your command.
moat:   fix the command above, or remove hooks.pre_run from moat.yaml.
```

Same exit code, but the user now sees it was the pre_run hook, the exact command, and how to fix it.

## How

`run_pre_run_hook` captures the hook's exit status (with `set -e` briefly off so it can report instead of aborting), and on non-zero prints a framed `moat: pre_run hook failed …` block to stderr and exits with the hook's status. Success/absent-hook behavior is unchanged.

## Test (red-green)

`TestMoatInitPreRunHookBehavior` extracts the real `run_pre_run_hook` from the embedded `moat-init.sh` and executes it under `sh` with `set -e`, asserting:
- failing hook → exit code 42 **and** the framed message, and the entrypoint does **not** continue;
- successful hook → continues, no failure message;
- absent hook → no-op.

Verified red→green: the failing-hook subtest fails on the pre-fix script (`missing framed failure message`, output is just the bare hook output) and passes after. The exit-code assertion already matched pre-fix (`set -e` propagates it), so the test isolates the framing — the actual fix. Skips on Windows/root (non-root branch). Also verified end-to-end with a real container in the dev environment.

## Scope

The host-side wrapper still prints the generic `Error: run failed: exit code N` (it can't distinguish hook-vs-command), but the container-side framing is now clear in the output / `moat logs`. A host-side "this run had a pre_run hook" hint could be a follow-up.


dpup:
Thanks — good review. Added the comment on both branches explaining the subshell asymmetry (non-root wraps in a subshell so the hook's `cd` can't move the entrypoint's cwd; the gosu branch is already a separate process) in f611651. The root/gosu-branch behavioral coverage is noted as a follow-up for when a rootful CI env is available; the framing path is shared so the risk is low.

Issue #348, opened 2026-04-29
Proxy daemon does not register MCP servers for runs with `mcp[].auth.grant` (post-#338   regression in 0.5.1)


PatrickSpieker:
## Summary

Following up on #335, which was fixed by #338. With Moat **0.5.1** (built 2026-04-28, includes #338), the
original failure mode is gone — credential loading no longer 500s — but a *new* failure replaces it: the
proxy daemon never registers runs that declare MCP servers, so the relay endpoint serves `404` for every
request.

The agent's `~/.claude.json` looks correct, the run metadata correctly includes the auto-included `mcp-*`
grants from `appendMCPGrants()`, but `~/.moat/proxy/runs.json` is never updated to include the new run,
and the relay can't route to it.

This appears to be a regression in the run-registration path, distinct from the credential-loading path
#338 fixed. Symptoms are similar enough to #335 that it could easily be mistaken for the same bug, but
the error wording and diagnostic evidence are different.

## Environment

```
$ moat version
moat 0.5.1
commit: fc0596858df7275a341e3ca195860cd773c4e564
built:  2026-04-28T22:14:27Z
go:     go1.25.5

Runtime: apple containers (macOS, arm64)
```

Reproduces with both Render MCP (`mcp.render.com/mcp`) and Linear MCP (`mcp.linear.app/mcp`) — same as
the original #335 repro patient.

## Steps to reproduce

1. Store credentials for both MCP servers (one-time):
 ```bash
 moat grant mcp render
 moat grant mcp linear
 ```

2. Use this `moat.yaml` (minimal — adapted from a real project's config; the `runtime: apple` and
Playwright bits aren't load-bearing for this bug):

 ```yaml
 runtime: apple
 dependencies:
   - python
 grants:
   - claude

 mcp:
   - name: render
     url: https://mcp.render.com/mcp
     auth:
       grant: mcp-render
       header: Authorization
   - name: linear
     url: https://mcp.linear.app/mcp
     auth:
       grant: mcp-linear
       header: Authorization
 ```

3. Start a fresh run:
 ```bash
 moat claude .
 ```

4. From inside the container, inspect the generated relay config:
 ```bash
 python3 -c '
 import json
 cfg = json.load(open("/home/moatuser/.claude.json"))
 for name in ("render", "linear"):
     print(name, json.dumps(cfg["mcpServers"][name]))
 '
 ```

 Output (looks correct):
 ```
[...]

PR #378, opened 2026-06-16
test(daemon): regression test for MCP relay token-path routing (#348)

dpup:
## What

Adds a regression test for the relay-routing contract behind #348.

moat builds MCP relay URLs as `/mcp/{token}/{server}` (`internal/run/manager.go`), and the credential proxy must route them to the per-run MCP server resolved from `{token}` — not treat the token as the server name. In 0.5.1 the token-aware relay path wasn't taken, so requests fell through to the plain relay, which read the first path segment (the token) as the server name and 404'd:

```
MOAT: MCP server '<token>' not configured. Available servers: N. Check moat.yaml.
```

## Status of the bug

**Already fixed on `main`** — not a code change here, just a guard. Investigation found:

- The daemon wires `proxy.SetContextResolver(...)` (`cmd/moat/cli/daemon.go`), which enables gatekeeper's token-aware `handleDirectMCPRelay` (present in the pinned **gatekeeper v0.2.0**). 0.5.1 predated that relay path.
- The run-context chain carries MCP servers end to end: `req.MCPServers` → `ToRunContext` → `ToProxyContextData` → resolver.

Verified in pure Go: with the resolver wired, `POST /mcp/{token}/render` returns **200** (relayed); without it, the request reproduces the issue's 404 byte-for-byte.

## The test

`TestMCPRelayTokenPathRouting` drives moat's real `RunContext.ToProxyContextData()` through the gatekeeper proxy with a token-in-path request, asserting it relays to the upstream (200) rather than 404ing. It guards both moat's run-context wiring and the relay-routing contract. Pure HTTP — **no container runtime required**, so it runs in normal unit CI.

## Not closing the issue yet

Using `Refs #348` (not `Closes`) deliberately: #348 was reported against Apple containers / live MCP servers, so before closing we want the original reporter to confirm they can no longer reproduce on a current build. Asking on the issue.

Refs #348

### DIFF
--- a/internal/daemon/mcp_relay_test.go
+++ b/internal/daemon/mcp_relay_test.go
@@ -103,6 +103,8 @@ func TestMCPRelayTokenPathRouting(t *testing.T) {
 			t.Fatalf("expected 404 without a context resolver, got %d\nbody: %s", res.StatusCode, strings.TrimSpace(string(body)))
 		}
 		// The token (first path segment) is what got looked up as the server name.
+		// Note: this asserts gatekeeper's current 404 body shape; if a gatekeeper
+		// update changes the relay's error wording, update this expectation.
 		if !strings.Contains(string(body), token) {
 			t.Fatalf("expected the token to be misread as the server name in the 404 body\nbody: %s", strings.TrimSpace(string(body)))
 		}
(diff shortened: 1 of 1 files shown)


dpup:
Both addressed in a0156a7:
- Token constant is now 64 hex chars (32 bytes), matching a real ProxyAuthToken.
- Added a `without a resolver` sub-test that asserts the same token-path request 404s with the token misread as the server name — the explicit red half, so the guard demonstrably fails on the broken path rather than passing incidentally.

Thanks for the review.

PR #380, opened 2026-06-17
fix(github): surface re-grant hint when an injected credential is rejected

### LINKED ISSUE #374: github: stale/invalid stored token fails with opaque `could not read Username` — surface a re-grant hint

dpup:
## Summary

Fixes #374. When a stored GitHub token is expired/invalid, HTTPS git failed with only git's opaque `fatal: could not read Username for 'https://github.com'` — the proxy's actual `401 invalid credentials` was buried in the run's network log, so the error read like a code bug rather than a stale credential.

After a non-interactive run completes, moat now scans the run's network log and, when a `github`-granted request to `github.com`/`api.github.com` returned 401/403, prints an actionable hint:

> ⚠ GitHub rejected the injected credential — the stored token may be expired. Run `moat grant github` to refresh it.

## Design

- Reuses the existing pattern from `doctor claude` (which already scans the network log for Anthropic 401s) and the existing `RunStore.ReadNetworkRequests()` infra.
- The mapping is a small `grantAuths` table (grant → hosts → hint), so other grants can be added later. v1 covers `github` (the reported case).
- Host matching parses the URL host and suffix-matches, so look-alikes like `notgithub.com` don't trigger.
- Purely additive: only prints when a real 401/403 to a granted host is found.

## Scope / non-goals

This makes the failure *legible*; it does not auto-refresh the token. It's also independent of #373 (the daemon serving a stale credential after re-grant) — that's a separate root cause.

## Testing

`credentialRejectionHints` is unit-tested (github 401, api.github.com 403, not-granted, 200 success, no-requests, look-alike host). `make lint` clean; build green.


dpup:
Thanks — all five addressed in 5578cd8:

1. **Spurious warning on success** — the hint is now gated on a failed run (`err != nil`), *and* `hasUnrecoveredRejection` only flags a host whose 401/403 had no subsequent 2xx (so retry / SSH-fallback recovery is suppressed).
2. **403 wording** — now "expired or missing a required scope".
3. **CHANGELOG** — links to the PR (#380).
4. **Swallowed read error** — now logged at debug.
5. **Mixed-result test** — added `401 then 200 (recovered) → no hint` and `api 403 unrecovered while git 200` cases.

dpup:
Good catch — fixed in 4c32243. `hasUnrecoveredRejection` now processes requests in chronological order and a later rejection un-recovers a prior success (`delete(succeeded, host)`), so `[200, 401]` flags while `[401, 200]` and `[200, 401, 200]` do not. Added both ordering test cases.

Self-reviewed the full change before pushing — the recovery matrix is now: `[401]`→hint, `[200]`→none, `[401,200]`→none, `[200,401]`→hint, `[200,401,200]`→none, api-403-while-git-200→hint (per-host), not-granted→none, look-alike host→none. Also rebased onto main now that #381 is merged (resolved the CHANGELOG `### Fixed` overlap).

PR #381, opened 2026-06-17
fix(github): stop proactively refreshing env-sourced tokens so re-grant takes effect (#373)

### LINKED ISSUE #373: daemon serves a stale credential after re-grant until `moat proxy stop`

dpup:
## Summary

Fixes #373. After `moat grant github` updated a stored credential, an already-running proxy daemon kept injecting the **previous** token until `moat proxy stop`.

## Root cause

The daemon's background token-refresh (`internal/daemon/refresh.go`) runs an initial refresh at registration and then every 5 minutes. For GitHub, `Refresh` re-derives the token from its *source* — `gh auth token` or the daemon's `GITHUB_TOKEN`/`GH_TOKEN` env — and writes the result back to both the proxy and the credential store.

For an **env-sourced** token, that source is the **daemon's own process environment, frozen at spawn**. So re-deriving can only ever produce a value that's stale relative to a later `moat grant` — and it silently overwrites the freshly-granted token (in the store too). `moat proxy stop` "fixes" it only because a new daemon inherits the current shell's env.

## Fix

`CanRefresh` now returns true only for the **`gh auth token` (CLI) source**, not the env source:

- **`SourceCLI`** — the gh keyring is shared state that `moat grant github` itself reads, so the daemon re-deriving from it never diverges from a re-grant, and long runs still follow gh's own token rotation. **Kept.**
- **`SourceEnv`** — frozen in the daemon, can only go stale vs a re-grant. **Treated as static** (no proactive refresh); a re-grant takes effect on the next run, and a stale token now surfaces as a 401 with a re-grant hint (#374) rather than being silently re-injected.

## Why scoped to github (and to env)

Other refreshable providers (`claude`, `gemini`, `oauth`, `meta`) refresh by *minting a new token* (OAuth refresh token / app-secret exchange) — they always move forward and never re-read a side-channel that `moat grant` can't update, so they're unaffected. github's env source is the only path that can move backward, so the fix is targeted there. (A blanket "skip refresh unless near expiry" rule was considered and rejected — it would break `meta`, whose initial short-lived→long-lived exchange runs with no recorded expiry.)

## Testing

`TestProvider_CanRefresh` updated: env source now expects `false`. github + daemon suites pass under `-race`; `make lint` clean. The `Refresh` method's env branch is retained (still valid and exercised when called directly) — only the *proactive* daemon path is gated off.

Independent of #374 (the 401 hint), which makes the residual staleness legible.


dpup:
Thanks — fixed the stale `SourceEnv` constant comment in 7389e36 (now '- static (not proactively refreshed; see CanRefresh)'), and clarified the `SourceCLI` one too. Also linked the changelog entry to the PR per CLAUDE.md. Agreed on the vestigial `Refresh` env branch — leaving it for the direct-call tests / a possible future reactive path.

PR #387, opened 2026-06-17
fix(mcp): address proxy relay via moat-proxy so the token is stripped

dpup:
## Problem

Remote MCP servers fail to connect with `MOAT: MCP server '<64-hex>' not configured` (HTTP 404). The `<64-hex>` "server name" is actually the run's proxy auth token.

I traced it in a real run's logs:

```
Streamable HTTP error: Error POSTing to endpoint:
MOAT: MCP server '9b9d…token…d92b' not configured
```

## Root cause

The proxy has two MCP entry points:
- `handleDirectMCPRelay` — for **direct** requests (`r.URL.Host == ""`). Strips the token from `/mcp/{token}/{name}` and resolves the run. ✅
- `handleMCPRelay` — for **proxied** requests. Treats the first path segment as the server name, so `/mcp/{token}/{name}` → server name = `{token}` → 404. ❌

The relay URL written into the container's `.claude.json` (`internal/run/manager.go`) addressed the proxy by the raw host-gateway IP (`GetHostAddress`), which is **not** in `NO_PROXY` — only the synthetic `moat-proxy` host is. So the MCP client's request was routed *through* the CONNECT tunnel, arrived as a proxied request (`r.URL.Host` set), and fell into the wrong handler.

## Fix

Address the proxy by the synthetic `moat-proxy` host — the only host in `NO_PROXY` — so the client connects directly and the proxy strips the token via `handleDirectMCPRelay`. This mirrors the `ANTHROPIC_BASE_URL` relay, which already used the synthetic host.

```go
-proxyAddr := fmt.Sprintf("%s:%d", m.defaultRuntime().GetHostAddress(), r.ProxyPort)
+proxyAddr := fmt.Sprintf("%s:%d", syntheticProxyHost, r.ProxyPort)
```

## When it regressed

#321 (v0.5.1, 2026-04-16) moved the proxy to synthetic hostnames and narrowed `NO_PROXY` from `<gateway-ip>,localhost,127.0.0.1,buildkit` to `moat-proxy,…`. Before that, `GetHostAddress` (the gateway IP) was itself in `NO_PROXY`, so the relay URL connected directly. #321 updated the `ANTHROPIC_BASE_URL` relay to the synthetic host but missed the MCP relay. Affects all remote `mcp:` servers.

## Testing

- `make lint` — 0 issues
- `go test -race ./internal/run/ ./internal/daemon/` — the daemon MCP token-path regression test (#348) passes; one pre-existing unrelated failure (`TestManagerExecInteractive_RuntimeResolveError`) fails identically on `main`.
- Note: the relay URL is baked into `.claude.json` at container creation, so existing containers must be recreated to pick up the fix. Verify by starting a fresh run in `examples/mcp-notion` and checking `/mcp` in Claude Code.

Also updates `examples/mcp-notion` to demonstrate multiple MCP servers (linear, posthog, notion).


dpup:
Thanks — both addressed in 0749058.

**1. Example grants** — added `oauth:linear` and `oauth:posthog` and updated the header comment to grant all three. One correction on the diagnosis: this wouldn't be a *silent* auth failure. `appendMCPGrants` auto-applies each `mcp:` server's grant to the run (so credentials are injected even when the grant isn't duplicated under `grants:`), and `validateMCPGrants` fails the run pre-start with an actionable `moat grant …` message if a credential is missing. The real gap was the **header comment** only mentioning `moat grant oauth notion`, so a user following it would hit that pre-run error for linear/posthog. Listed the grants explicitly anyway since this is a teaching example.

**2. Inline comment** — converted to a block comment above the line.

PR #400, opened 2026-06-23
fix(claude): keep bypass-permissions mode across the fullscreen-renderer switch

abezzub-dr:
## Problem

In `moat claude` sessions, bypass-permissions mode silently turns off mid-session. After accepting Claude Code's "Try the new fullscreen renderer?" prompt, Claude starts prompting for tool calls it shouldn't (e.g. a plain `cat`), and the `⏵⏵ bypass permissions on` indicator disappears.

## Root cause

moat enables bypass mode **only** via the `--dangerously-skip-permissions` CLI flag. When the user accepts the fullscreen-renderer upsell (or Claude silently auto-graduates the session), Claude Code **re-execs its own process with an empty argv**, dropping the original CLI flags. Bypass mode reverts to the settings default — and `settings.json` only had `skipDangerousModePermissionPrompt` (which suppresses the warning banner, not the mode).

Verified against the live container (PID 1 cmdline became bare `claude`) and the Claude binary: the renderer is chosen by `switch(settings.tui)`, and the upsell gate `_uc()` returns false when `settings.tui` is already set.

## Fix

Two complementary changes to the generated `settings.json`:

1. **Pin the renderer** via `tui` — mirrored from the host's `~/.claude/settings.json` choice, defaulting to the classic renderer (`"default"`). Setting `tui` suppresses the upsell and the silent auto-graduation, so the flag-dropping re-exec never fires (this also protects `--resume`/`--continue`/initial prompt, which the re-exec equally discards).
2. **Persist bypass mode** via `permissions.defaultMode: "bypassPermissions"` (when bypass is requested) so it survives any re-exec that does occur (e.g. a manual `/tui` switch). Existing permission sub-keys (allow/deny/ask from `~/.moat/claude/settings.json`) are preserved.

`--noyolo` opts out of bypass as before.

## Testing

- New unit tests with companion cases: `EnableBypassPermissionsMode` (set + preserve-existing-keys), `tui` marshal round-trip (set + empty-omitted), merge semantics (override-wins + base-kept + nil-clone), and host-`tui` mirroring through `LoadAllSettings` (mirror + skip-host).
- `go build ./...`, `go vet`, and `internal/providers/claude` tests (incl. `-race`) pass.

## Docs

- New "Terminal renderer" section in the Claude Code guide; permission-handling section notes the persisted mode; corrected the host-fields note to include `tui`.


dpup:
Thanks for this. Have you figured out how to get fullscreen working with scrollwheel?

abezzub-dr:
For me scrolling works in fullscreen mode with moat. Maybe there is another intermittent issue?

PR #402, opened 2026-06-23
feat: native (in-VM) volume support via volumes type: volume

abezzub-dr:
## Summary

Adds a `type` field to `moat.yaml` `volumes:` so a volume can be backed by a **Docker named volume** instead of a host bind mount.

- `type: volume` → `moat_<agent>_<name>` Docker named volume on the container engine's native filesystem (inside the VM where applicable). Not visible on the host; bypasses the host↔VM filesystem-sharing layer that bind mounts cross on VM-based runtimes. Faster for I/O- and metadata-heavy directories, and avoids that layer's differing file-locking / memory-mapping semantics.
- Default is unchanged (`bind`) — fully backwards-compatible; existing configs behave identically.
- **Permission fix:** Docker creates named-volume roots `root:root`. They're chowned to the run user by whichever path applies — a root entrypoint chowns via `moat-init` (driven by `MOAT_VOLUME_CHOWN`); a non-root container (e.g. Linux running as the workspace UID) chowns via a short root helper container before start. Named volumes also force the `moat-init` entrypoint, so a custom image can't end up running as the non-root user with no chown.
- **Docker-only:** the Apple `container` runtime rejects `type: volume` via a single shared validator used at both config-load (explicit `runtime: apple`) and run-setup (auto-detected). The agent name is validated against the Docker volume-name charset when `type: volume` is used.
- Daemon API untouched (volume setup is CLI-side).

## Test plan

- [x] Unit: config validation both directions (valid/invalid `type`, `type: volume` on Apple, whitespace target, invalid agent name); `buildContainerMounts` (bind → `TypeBind`, `Volume` → `TypeVolume`); `volumeMount` / `volumeOwnershipPlan` / `volumeOwnershipHelperConfig`; `needsInit` (named volumes → moat-init required).
- [x] `go build ./...` clean; `sh -n moat-init.sh` clean; no new lint findings.
- [x] **macOS / Docker Desktop (root entrypoint):** cold `type: volume` run verified end-to-end — caches on native ext4, a ~1300-package `pnpm install` with no SQLite-WAL hang, host repo untouched, `node_modules` writable.
- [x] **Helper chown** verified against a real moat-built image — the entrypoint override chowns the volume root to a non-root UID, whereas a `Cmd`-only form fails with EPERM (regression-guarded by a unit test).
- [ ] **Linux non-root container, full run:** not exercised end-to-end; covered by the unit decision logic plus the manual helper-chown check above.

## Follow-up

`moat volumes ls/rm/prune` still manage only `type: bind` volumes (host dirs under `~/.moat/volumes/`); they don't yet list or remove `type: volume` Docker named volumes (`moat_<agent>_<name>`). Until that lands, the reference docs point users to `docker volume rm` for those. Extending the CLI to manage named volumes (with a type column and in-use handling) is intended as a follow-up.


abezzub-dr:
@dpup thoughts on this feature? It makes dependency install on my project go from 5 minutes down to 30 seconds on fresh container start.

dpup:
I have a branch that implements basically the same feature! Let me have Claude compare branches.

dpup:
Yours has some things that are better than mine, so I'll merge and integrate on top.

PR #404, opened 2026-06-24
feat: volume-mode workspaces

dpup:
## What

Adds an opt-in `workspace.mode: volume` that copies the host working tree into an **ephemeral per-run Docker named volume** mounted at `/workspace`, instead of bind-mounting the host tree. Motivation: protect the host tree from agent writes, and get native-speed I/O on macOS (where bind mounts are slow). Changes are recovered to the host via snapshots.

```yaml
# moat.yaml
workspace:
  mode: volume          # or: moat run --workspace-mode volume
  # excludes are carried on a /workspace mount entry:
  # mounts: [{ source: ".", target: /workspace, exclude: [node_modules, dist/sub] }]
```

Docker-only; Apple containers and git worktrees/submodules are rejected with a clear error (use `workspace.mode: bind` for those).

## How it works

- The run manager mounts the host tree **read-only** at a staging path (`/mnt/host-workspace`) and a fresh per-run named volume (`moat-ws-<run-id>`) at `/workspace`.
- `moat-init.sh` (forced to run as root) copies staging → volume honoring `exclude:` patterns, then chowns `/workspace` to the run user before dropping privileges via gosu.
- Extraction reuses the existing **archive** snapshot backend: `moat snapshot <run>` materializes the volume to a host temp dir via a short-lived sidecar `tar` container, then archives it with `.git` included. Recover with `moat snapshot restore <run> --to <dir>`.
- `moat destroy` / `moat clean` refuse to delete a volume run that has no extraction snapshot unless `--force` (the volume is the only copy of the agent's work).
- The proxy daemon's idle cleanup GCs orphaned `moat-ws-*` volumes whose run directory is gone.

Built on top of #402's named-volume primitive (`MountConfig.Volume`); this PR adds the workspace-copy layer, snapshot extraction, the data-loss guards, and the GC.

## Review

Self-reviewed with `/ce-code-review` (multi-persona) before pushing. Findings were fixed across three follow-up commits, each verified by unit tests + the Docker e2e lifecycle:

- **Data integrity:** `VolumeExport` now captures the source-tar exit status (a POSIX pipeline previously masked a partial-but-valid source archive → silently truncated snapshot). Orphan GC lists volumes before snapshotting live run dirs (closes a TOCTOU that could reclaim a brand-new run's volume).
- **Ownership:** volume mode's `0:0` entrypoint is now treated as a root entrypoint in both the chown-env and ownership-helper paths, so combining volume mode with a config `type: volume` volume chowns it to the run user (not root).
- **Guards:** `hasExtractionSnapshot` reads the snapshot store directly so the destroy/clean guard works even after the host workspace is gone; `moat clean` exits non-zero when volume runs are skipped (machine-readable signal). `VolumeExport` is time-bounded.

## Testing

- `make test-unit` (race) green across config/container/run/daemon/snapshot/cli, plus new companion tests (chown routing, `ListSnapshots`, `hasExtractionSnapshot`, GC live-set, `IsVolumeMode`).
- `internal/e2e/volume_workspace_test.go` — Docker-gated full lifecycle (create → agent commits in volume → snapshot → restore → destroy guard) passes.
- `make lint` 0 issues.

## Docs

- New guide: `docs/content/guides/15-volume-workspaces.md`
- `reference/01-cli.md` (`--workspace-mode`), `reference/02-moat-yaml.md` (`workspace.mode`)
- Design + plan in `docs/plans/2026-06-23-volume-mode-workspaces-{design,plan}.md`


dpup:
Addressed the review-bot findings in `11af7ae`:

- **`.mcp.json` clobbered in volume mode** (real bug) — moat-init wrote Codex/Gemini `.mcp.json` before `populate_workspace_volume` extracted the staging tree, so a project's own `.mcp.json` could overwrite moat's. Moved the copy into `setup_workspace_mcp_json` and call it **after** populate; moat's config now wins in both modes.
- **`clean --force` overloading** — split it: `--force` now only skips the confirmation prompt; a new **`--force-volumes`** is required to remove volume-mode runs lacking an extraction snapshot. Added `shouldSkipVolumeRunForClean` + a companion test, and updated the CLI reference + volume guide.
- **`clean` double error message** — `cleanCmd` sets `SilenceErrors`, so the summary line and the non-zero-exit error no longer both print.
- **`listSnapshots` backend** — forces the archive backend for volume runs (matching prune/restore) so the footer doesn't print `apfs` on macOS.
- **Stale `--no-dereference` comments** in `moat_init_volume_test.go` — corrected; the script does not pass the flag (GNU tar 1.34; symlink-preserve is the default), as `TestMoatInitNoNoDeref` asserts.

Lower-priority items left as follow-ups (noted, not blocking): `VolumeList` interface-doc vs hidden label filter; `GuardVolumeWorkspace` wording for non-git dirs; `VolumeExport` hardcoded `debian:bookworm-slim`; `list`/`prune`/`restore` still requiring the workspace dir to exist; volume-mode snapshot *triggers* targeting the host staging dir.

All unit tests (race) + the Docker e2e lifecycle pass; lint clean.

dpup:
Addressed the latest review in `81a928f`: **`moat snapshot {list,prune,restore --to}` no longer fail when the original workspace dir is gone** (they only read `snapshotDir`) — new `snapshotEngineWorkspace` falls back to a guaranteed-existing placeholder; in-place restore still requires the real path. Minors: `VolumeExport` names the base image in its create-error; `VolumeList` interface doc now notes the `moat=workspace` label filter. Companion test added; unit (race) + Docker e2e + lint all green.

dpup:
Fixed in `26f6eaf`: the **interactive keyboard snapshot now refuses volume-mode runs** (it used the in-process `SnapEngine`, which points at the host staging dir — a keyboard snapshot would archive the wrong tree yet still satisfy `hasExtractionSnapshot` and unlock destroy). It now directs the user to `moat snapshot <run>`. Companion test added (volume → no snapshot; bind → still snapshots). Minor: `snapshots.json` now written `0600` to match the archives. Unit (race) + e2e + lint green.

Issue #408, opened 2026-06-26
Routing proxy: actionable error when configured port is busy + redundant daemon proxy


dpup:
Surfaced while working on #407 (routing discovery index + `moat open`).

## 1. Actionable error when the configured routing port is busy

`routing.Lifecycle.EnsureRunning` binds `globalCfg.Proxy.Port` (default `8080`) and returns a raw error if it's taken:

```
Error: creating run: starting routing proxy: starting proxy: listen tcp 127.0.0.1:8080: bind: address already in use
```

A run that exposes `ports:` won't start if 8080 is busy, and the message doesn't tell the user what to do.

**Design decision (dpup): keep the port deterministic — do NOT auto-fall-back to a random port.** Advertised endpoint URLs (`moat open`, the discovery index, `MOAT_URL`) must stay consistent/bookmarkable run-to-run, so a random fallback is undesirable. The port is already configurable via `MOAT_PROXY_PORT=<port>` or `proxy.port` in global config.

### Proposed fix
Catch `EADDRINUSE` in `EnsureRunning` and return an actionable error — e.g. "routing port 8080 is in use; set a different port with `MOAT_PROXY_PORT=<port>` or `proxy.port` in ~/.moat/config" — instead of the raw bind error. No fallback, no lock/mismatch changes needed.

## 2. Redundant daemon routing proxy

`cmd/moat/cli/daemon.go` (~L207-220) starts a *second* routing proxy via `routing.NewProxyServer` + `Start(0)` on a random port, serving the same `routes.json` as the per-agent `Lifecycle` proxy that `moat run` actually advertises. It doesn't write the proxy lock, so `moat open` / `moat proxy status` never see it. It looks redundant/dead relative to the lifecycle proxy and caused confusion during #407 testing. Worth confirming whether it has a purpose (e.g. host-network runs) and removing it if not.

dpup:
Item 1 (actionable busy-port error, no fallback — keeping the port deterministic per the design note) is addressed in #409. Item 2 (redundant daemon routing proxy) remains open.

PR #409, opened 2026-06-26
fix(routing): actionable error when the routing port is in use

dpup:
## What

When the routing proxy's port is already in use, `moat run` (for an agent that exposes `ports:`) now fails with an actionable message instead of a raw bind error:

**Before**
```
Error: creating run: starting routing proxy: starting proxy: listen tcp 127.0.0.1:8080: bind: address already in use
```

**After**
```
Error: creating run: starting routing proxy: routing proxy port 8080 is already in use — choose another with MOAT_PROXY_PORT=<port> or set proxy.port in ~/.moat/config.yaml: ...
```

## Why

The routing proxy uses a **deterministic** port (default 8080, configurable via `MOAT_PROXY_PORT` or `proxy.port`) and intentionally does **not** fall back to a random port — advertised endpoint URLs (`moat open`, the discovery index, `MOAT_URL`) must stay stable and bookmarkable run-to-run. So the right behavior when the port is busy is to tell the user how to pick a different one, not to silently move.

Surfaced while working on the routing discovery index / `moat open` work (#407). Addresses item 1 of #408.

## Implementation

`routing.Lifecycle.EnsureRunning`: on `EADDRINUSE` from the listener, wrap the error with the port number and the two ways to change it. Other bind errors keep the existing generic message.

## Testing

- New `TestEnsureRunningPortInUse`: occupies a port, then asserts `EnsureRunning` returns an error naming the port and `MOAT_PROXY_PORT`.
- `gofmt` / `go vet` / `golangci-lint` (0 issues) / `go test ./internal/routing/` all green.


dpup:
Thanks — both notes considered, leaving the code as-is:

- **Port-0 message degradation:** the `EADDRINUSE` branch is actually unreachable when `desiredPort=0`, since `Start(0)` asks the OS for a free port and never collides — so `lc.port` is the real (non-zero) configured port whenever this message fires. The confusing "port 0" text can't occur in practice.
- **`syscall.EADDRINUSE` POSIX-only:** consistent with moat's existing POSIX-only assumptions (Unix-domain `daemon.sock`, `process.Signal(syscall.Signal(0))` liveness), and CI builds on Linux. If a Windows target ever appears it'd need build-tagged errno handling across the codebase, not just here — out of scope for this fix.

dpup:
Re-checking this: two review passes ran here and I only replied to the second. The **first** pass raised two points I missed — (1) the message still wrapped the raw `bind: address already in use` via `%w`, re-introducing the noise it was meant to remove, and (2) the test didn't assert the `proxy.port` remedy. Both are now fixed in #411 (drop the wrap + tighten the test). Thanks for the nudge.

PR #416, opened 2026-06-26
chore(security): upgrade dependencies to clear called vulnerabilities

dpup:
While probing the tooling-floor work (`govulncheck`), I found the codebase **calls 36 known vulnerabilities** (govulncheck's reachable set, not just "present in the tree"). This PR upgrades every dependency that has an available fix, clearing **15** of them.

## Upgrades (build + tests green)
| Module | From → To | CVEs cleared |
|--------|-----------|--------------|
| `containerd/v2` | 2.2.1 → 2.2.5 | 6 |
| `golang.org/x/crypto` | 0.49.0 → 0.53.0 | 2 |
| `in-toto-golang` | 0.9.0 → 0.11.0 | 1 |
| `go-git/v5` | 5.16.4 → 5.17.1 | 3 |
| `moby/buildkit` | 0.27.0 → 0.28.1 | 2 |
| `ulikunitz/xz` | 0.5.12 → 0.5.15 | 1 |

govulncheck's reachable count drops **36 → 21**.

## What's left (out of scope here)
- **`docker/docker`** (5) — "Fixed in: N/A" upstream; nothing to bump to yet.
- **Go standard library** (~16: `crypto/x509`, `crypto/tls`, `html/template`, `net/*`, `archive/tar`, …) — cleared by building with a newer Go toolchain (1.25.6 → 1.25.11), a separate decision since it raises the build-requirement floor for all contributors.

Because those remain, a *blocking* `govulncheck` CI gate would fail today, so I haven't added one — that lands once docker/stdlib are addressable.

## Also
- Added a CI **`go mod tidy` drift gate** so `go.mod`/`go.sum` staying untidy fails fast (complements the dependency hygiene this PR is about).

## Testing
`go build ./...`, `go vet ./...`, and the full non-network test suite pass (the pre-existing `deps` GitHub-URL-existence test fails only in this sandbox's blocked network; it passes in CI). No source changes — pure dependency + CI config.


dpup:
On the version nit: the two numbers refer to different things — `go.mod`'s `go 1.25.5` is the *language directive*, while govulncheck reports stdlib findings against the *build toolchain* it ran under (go1.25.6 here). Both are below the patched stdlib versions (1.25.10/1.25.11), so the substance holds: the stdlib findings clear once built with a newer toolchain. The CHANGELOG entry deliberately omits a specific number to avoid that confusion.

PR #418, opened 2026-06-27
chore(lint): enable exhaustive and testifylint

dpup:
From the craftsmanship audit's tooling-floor move — raise the lint floor with two checks that catch real issue classes, and fix what they flag. **+23/-5 lines.**

## `exhaustive` (with `default-signifies-exhaustive: true`)
Enum switches must handle every case or have a default. Configured so a `default:` counts as exhaustive — flagging only switches that silently skip cases with **no default** (the real silent-gap risk).

Of the 19 raw findings, 14 already have a default. The 5 without one are made explicit (all behavior-preserving — the unhandled cases already did nothing):

| Switch | Fix |
|--------|-----|
| `exec.go` `term.EscapeAction` | explicit `case EscapeNone, EscapeStop:` no-op (small enum → future-proof) |
| `dockerfile.go` `spec.Type` categorization | documented `default:` |
| `script.go` `spec.Type` categorization | documented `default:` |
| `dockerfile.go` dynamic branch | `//nolint:exhaustive` — gated by `dep.IsDynamic()`, only `TypeDynamic*` reach it |
| `writer.go` `controlSeqKind` | `//nolint:exhaustive` — `ctrlNone` is gated out by the caller |

## `testifylint`
Idiomatic testify assertions. Auto-fixed `assert.Empty`/`Len`/`Positive` uses; switched one `assert.Error` → `require.Error` so the following `err.Error()` can't nil-panic if no error occurred (the exact bug the check exists for).

## Deferred
`noctx` (~66 findings — threading context through that many call sites is its own large effort) and `gofumpt` (repo-wide reformat — better as its own focused PR).

## Testing
`go build ./...`, `golangci-lint run ./...` (**0 issues** with both linters now enabled), and the affected package tests pass. No behavior change → no CHANGELOG.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -25,8 +25,15 @@ linters:
     - unused        # Unused code detection
     - bodyclose     # HTTP response body must be closed
     - prealloc      # Slice preallocation for performance
+    - exhaustive    # Enum switches must handle every case (or have a default)
+    - testifylint   # Idiomatic testify assertions (require for errors, Len/Empty, etc.)
 
   settings:
+    exhaustive:
+      # A switch with a default is treated as exhaustive; we only flag enum
+      # switches that silently skip unhandled cases with no default at all.
+      default-signifies-exhaustive: true
+
     errcheck:
       check-type-assertions: true
       check-blank: false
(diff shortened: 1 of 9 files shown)


dpup:
Thanks — acted on the one substantive observation: `writer.go`'s `ctrlNone` is now an explicit `case ctrlNone: return nil` rather than a `//nolint:exhaustive` suppression (`74e...` → see latest commit). That drops the suppression, documents the gated contract right at the case, and keeps `exhaustive` able to flag a future `controlSeqKind` addition. Lint stays at 0 issues without the nolint.

On the others, no change needed and I agree with your read:
- `require.Error` fix and the explicit `EscapeNone/EscapeStop` case — the two real improvements, glad they landed.
- `dockerfile.go` dynamic-branch `//nolint` — kept; the `IsDynamic()` gate makes non-dynamic cases dead code, so a nolint-with-reason is the honest call there (unlike `ctrlNone`, where an explicit case is natural).
- `assert.Positive` vs `assert.Greater` — equivalent for ints; took testifylint's idiomatic preference.

PR #419, opened 2026-06-27
build: adopt gofumpt as the canonical formatter

dpup:
From the craftsmanship audit's tooling-floor move — uniform, stricter-than-gofmt formatting enforced at every layer (editor hook → lint → CI). Two commits for easy review: the **policy change** (4 files) and the **bulk reformat** (87 files, pure formatting).

## Policy (commit 1)
- **`.golangci.yml`**: configured formatter `gofmt` → `gofumpt` (stricter superset). `golangci-lint run` runs the formatters and fails on any unformatted file — verified it flags a violation and exits 1 — so the manual `gofmt -l` CI step is **removed as redundant**.
- **Claude hook** (`.claude/hooks/go-fmt.py`): format edited `.go` files with `gofumpt` when installed, falling back to `gofmt` otherwise, so **editor-time formatting matches what CI enforces**. Added a bats case asserting the gofumpt-specific rule (blank-after-brace removal) — version-independent, unlike the `0o` octal rule.

## Reformat (commit 2)
`golangci-lint fmt` across the tree: `0644`→`0o644`, grouped `var (...)` blocks, no blank lines at block edges, etc. **Pure formatting, no behavior change.**

## Testing
`go build ./...`, `go vet ./...`, `golangci-lint run ./...` (**0 issues**, now gofumpt-enforced), `bats .claude/hooks/go-fmt.bats` (**6/6**), and the test suite all pass. The only non-`.go` changes are the 4 policy files. No CHANGELOG (internal formatting/tooling).

### DIFF
--- a/.claude/hooks/go-fmt.bats
+++ b/.claude/hooks/go-fmt.bats
@@ -67,3 +67,17 @@ GO
   [ "${status:-0}" -eq 0 ]
   grep -q "binary data" "$TMPDIR/data.gob"
 }
+
+# --- gofumpt-specific behavior (when gofumpt is installed) ---
+
+@test "applies gofumpt's stricter rules when gofumpt is available" {
+  command -v gofumpt >/dev/null || skip "gofumpt not installed; hook falls back to gofmt"
+  # A blank line right after a function's opening brace: gofumpt removes it,
+  # plain gofmt leaves it. Independent of the module's Go version (unlike the
+  # 0o octal rule), so it works on this standalone temp file.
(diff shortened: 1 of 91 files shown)


dpup:
Good catch — the gofumpt bats test was skipping in CI (no standalone gofumpt on `$PATH`; golangci-lint bundles it internally). Added an `Install gofumpt` step (pinned `v0.9.2`) to the test job before the hook tests, so test 6 now actually exercises the gofumpt path in CI rather than skipping. "6/6" now holds in CI too.

PR #423, opened 2026-06-29
refactor(run): extract SSH agent setup from Create into setupSSHAgent

dpup:
## What

First phase extraction from the ~2,489-line `Create` (follow-up to #421 split and #422 credential consolidation). The SSH-agent-proxy wiring — ~145 lines that load SSH key mappings, connect the upstream agent, start a TCP or Unix-socket filtering proxy, and produce the container env/mounts — moves into a `setupSSHAgent` method.

`Create` drops from **~2,489 → ~2,353 lines**.

## Shape

```go
type sshAgentSetup struct {
    server *sshagent.Server
    env    []string
    mounts []container.MountConfig
}

func (m *Manager) setupSSHAgent(r *Run, opts Options, sshGrants []string,
    hostAddr string, openCredStore func() (*credential.FileStore, error)) (sshAgentSetup, error)
```

The caller computes `sshGrants` once (it also feeds the image spec later in `Create`), calls the method, and on success threads the result back:

```go
sshGrants := filterSSHGrants(opts.Grants)
var sshServer *sshagent.Server
sshSetup, sshErr := m.setupSSHAgent(r, opts, sshGrants, hostAddr, openCredStore)
if sshErr != nil {
    cleanupDaemonRun()
    return nil, sshErr
}
sshServer = sshSetup.server
proxyEnv = append(proxyEnv, sshSetup.env...)
mounts = append(mounts, sshSetup.mounts...)
```

`sshServer` stays a `Create`-scoped var (used at 24+ later cleanup sites + the audit-func wiring), assigned from the result.

## Behavior-preserving

Verified line-by-line — identical error messages, identical per-error-path partial-resource cleanup, identical env/mount/log output. The only intentional differences:

- `cleanupDaemonRun()` (rollback of resources allocated *before* SSH setup) moves out of the method to the single caller error path — the method only cleans up resources *it* created (stops its own server, closes the upstream agent, removes the socket dir), preserving the exact same order.
- `cleanupSSH(sshServer)` on the address-parse error path is inlined as `setup.server.Stop()` with the same `"failed to stop SSH agent during cleanup"` debug log.
- The caller's error var is named `sshErr` (not `err`) so it doesn't leak into `Create`'s top-level scope — using `err` there tripped 14 `govet` shadow warnings against the 24 later `if err := ...` blocks.

The passed-in `openCredStore` closure keeps the "open the credential store once" guarantee from #422 (the method doesn't open its own store).

## Verification

`go build ./...` ✓ · `go vet` ✓ · `golangci-lint run` → **0 issues** ✓ · `go test -race -count=1 ./internal/run/` ✓.

Part of the `manager.go` decomposition; further phase extractions and the cleanup-rollback-stack remain.


dpup:
Thanks — addressed the nit in the latest push: documented the zero-`sshAgentSetup{}`-on-error convention in the method doc comment (one place rather than four inline comments), so a reader sees up front why the error paths return a fresh zero literal instead of the partially-populated `setup`.

dpup:
Follow-up per review feedback: moved `setupSSHAgent` + `sshAgentSetup` into a dedicated **`manager_ssh.go`** (matches the `manager_*.go` convention from the split, and sets the file-per-phase precedent for future extractions), and added **`manager_ssh_test.go`** with unit tests for the guard branches the extraction unlocked — no grants, missing `SSH_AUTH_SOCK`, credential-store error, no SSH key mappings, and upstream-agent connect failure (faked `openCredStore` closure + a real temp `FileStore`, no live SSH agent needed). The server-start/env-mount success paths still need a real agent and stay e2e-only. `Create` is now 2,353 lines; all checks green locally.

PR #425, opened 2026-06-29
refactor(run): extract container resource-limit resolution from Create

dpup:
## What

Next phase extraction from `Create` (follows #421/#422/#423/#424). The memory/CPUs/DNS/ulimits computation moves into a pure `resolveResourceLimits` method in its own **`manager_resources.go`**.

`Create`: **~2,315 → ~2,288 lines.**

```go
func (m *Manager) resolveResourceLimits(cfg *config.Config) (memoryMB, cpus int, dns []string, ulimits []container.Ulimit)
```

Call site: `memoryMB, cpus, dns, ulimits := m.resolveResourceLimits(opts.Config)` (the four values are consumed only at container-config build, unchanged).

## Behavior-preserving

Identical extraction, ulimit sorting, and the Apple-container AI-agent 8 GB default (Apple's 1 GB system default is too low for the agent CLIs; Docker stays unlimited unless configured). Verified the substantive logic line-by-line against `main`; the only change is `opts.Config` → the `cfg` parameter.

## Tests

`manager_resources_test.go`:
- nil config → all zero
- config propagation + ulimit sort-by-name
- Apple + AI agent + no memory → 8 GB default (via an Apple-runtime stub)
- Apple + non-AI agent → no default
- Apple + explicit memory → explicit value kept over the default

## Verification

`go build ./...` ✓ · `go vet` ✓ · `golangci-lint run` → **0 issues** ✓ · `go test -race -count=1 ./internal/run/` ✓ (5 new tests pass).

### DIFF
--- a/internal/run/manager_create.go
+++ b/internal/run/manager_create.go
@@ -2117,35 +2117,8 @@ region = %s
 	buildkitEnv := computeBuildKitEnv(buildkitCfg.Enabled)
 	proxyEnv = append(proxyEnv, buildkitEnv...)
 
-	// Extract container resource limits from config (applies to both Docker and Apple).
-	// Priority: explicit moat.yaml > agent provider default > runtime fallback.
-	var memoryMB, cpus int
-	var dns []string
-	var ulimits []container.Ulimit
-	if opts.Config != nil {
-		memoryMB = opts.Config.Container.Memory
-		cpus = opts.Config.Container.CPUs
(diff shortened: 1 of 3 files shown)


dpup:
Added `TestResolveResourceLimits_DockerAINoDefault` — the companion to the Apple default, asserting Docker + AI agent + no explicit memory stays 0. Good catch on the untested branch.

PR #432, opened 2026-07-01
ci(review): let the review bot run go build/vet/test to verify findings

dpup:
## Problem

Every `claude-review` pass closes with some variant of:

> I could not run `go build`/`go vet`/`go test`/`gofmt` in this sandbox — all non-trivial Bash commands required interactive approval unavailable in this run. Everything above is from static reading of the diff.

So the bot recommends "confirm `make lint`/`make test-unit` pass" but never runs them itself. Two causes in `.github/workflows/claude-code-review.yml`:

1. The action's `--allowedTools` only permitted `mcp__github_inline_comment__…` and three `gh pr …` commands. Any `go`/`make`/`gofmt` command fell outside the allowlist, prompted for approval, and was denied in headless CI.
2. The job had no Go toolchain set up at all.

## Fix

- Add an `actions/setup-go@v6` step (`go-version-file: go.mod`, cache) mirroring the CI `lint`/`test` jobs.
- Extend `--allowedTools` with `Bash(go build:*)`, `Bash(go vet:*)`, `Bash(go test:*)`, `Bash(gofmt:*)`, `Bash(make:*)`, `Bash(golangci-lint:*)`.
- Tell the reviewer in the prompt that Go is available and it should run targeted builds/tests to verify findings (preferring targeted package tests over the full suite, which CI already runs).

## Safety

The job is already gated to same-repo PRs (`if: github.event.pull_request.head.repo.full_name == github.repository`), so executing a PR branch's test code is the same trust boundary as the existing `Test` CI job — no new exposure to fork PRs.

Because `pull_request` runs the workflow from the PR's own head for same-repo PRs, this PR's own review run will exercise the new config.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -20,6 +20,14 @@ jobs:
         with:
           fetch-depth: 1
 
+      # Set up Go so the reviewer can actually run go build/vet/test/gofmt to
+      # verify findings, instead of falling back to static-only analysis.
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+          cache: true
+
       - name: Collapse previous Claude reviews
         env:
           GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


dpup:
Note: the `claude-review` job on this PR **skipped its own review** by design — `claude-code-action` refuses to run when the PR modifies its own workflow file ("the workflow file must have identical content to the version on the default branch"), an anti-tampering guard so a PR can't rewrite the reviewer to run against itself.

So this change can't be self-verified here. What *is* confirmed from the job log: `actions/setup-go` set up Go 1.25.11 (cache hit, CGO/gcc ready), so the toolchain half works. The expanded `--allowedTools` will take effect on the next non-workflow PR after this merges — that's where you'll see the reviewer actually run `go build`/`go test` instead of reporting it couldn't.

PR #433, opened 2026-07-02
feat(pi): add Pi coding agent provider

dpup:
## Summary

Adds `moat pi` — run the [Pi coding agent](https://github.com/earendil-works/pi) in an isolated Moat container with transparent credential injection.

Pi is a bring-your-own-key agent with **no credential of its own**, so it reuses Moat's existing `anthropic` (x-api-key) and `openai` (Bearer) grants. v1 supports those two backends only; every other backend, plus a missing or ambiguous grant, **fails hard before a container is created**.

## How it works

- **Backend selection** (`resolvePiProvider`): `--provider` / `pi.provider` override, otherwise inferred from the single configured grant. Both grants without an override → hard error asking you to choose. Unsupported backend → hard error. All checks run before `Create`.
- **Credential injection**: delegated entirely to the existing anthropic/openai grant providers (Pi honors `HTTP_PROXY`, verified). Pi's own `ConfigureProxy`/`ContainerEnv` are no-ops; nothing secret touches the container filesystem.
- **Runtime context**: injected via Pi's `--append-system-prompt` pointing at a mounted staged file, so it augments — not clobbers — the user's own `AGENTS.md` / `CLAUDE.md`.
- **Guards**: a `moat pi` run forces `cfg.Agent = "pi"` and skips claude/codex/gemini staging (which would otherwise trip on the shared anthropic/openai grant).

## Verification

- Pi honors `HTTP_PROXY` (a logging proxy captured `CONNECT api.anthropic.com:443`).
- Full `moat pi` run (DIND): image built with `pi-cli`, container started, Pi launched `--provider anthropic` and reached `api.anthropic.com` through the proxy (401 on a placeholder key — the expected signal; a real granted key completes).
- Fail-hard paths (no grant / unsupported provider / provider-without-grant) exit non-zero before container creation.
- `go build ./...`, `make lint` (0 issues), and unit tests green.

## Deferred (future work)

- Other Pi backends (Gemini, Bedrock, …) — only anthropic/openai are wired up.
- `pi.mcp` — Pi core has no MCP configuration format to target.
- `pi.sync_logs` — session-log host-mounting is claude-only today.
- `--grant claude` (OAuth) — request-shape mismatch; see the design appendix.

Design + plan: `docs/plans/2026-07-01-pi-agent-provider-design.md` and `...-plan.md`.


dpup:
Thanks for the review — addressed the findings:

1. **`cli.go:95` — `--worktree` bypassing the worktree's `pi.provider`/`pi.model` (correctness):** Fixed. Added an optional `Preflight func(cfg *config.Config) error` hook to `ProviderRunConfig` that runs *after* workspace/worktree/config resolution and before grant assembly. Pi now resolves the backend inside `Preflight`, against the final (post-worktree) config, and fails hard there. This also let me delete the duplicated `loadWorkspaceConfig` workspace-parsing. (`ResolveWorktreeWorkspace` creates the worktree + prints feedback, so calling it a second time from `runPi` wasn't viable.)

2. **`init.go:70` — dead `pi` entry in `agentConfigs()`:** Removed. As you noted, Pi has no credential of its own, so `getCredGrant` could never win the auto-detect loop (claude/codex already claim anthropic/openai). Replaced with a comment explaining that a Pi project is scaffolded via `agent: pi` in `moat.yaml`.

3. **`agent.go:35` — vestigial `MOAT_PI_INIT`:** Removed. The context file is read directly from the mount via `--append-system-prompt`, and `moat-init.sh` has no Pi copy step, so the env var was unused.

4. **`cli.go:154` — `credentialConfigured` duplicating `codex`/`gemini` `HasCredential`:** Left as-is for this PR (non-blocking). A shared `internal/credential` helper would also mean refactoring codex/gemini's existing `HasCredential`, which is out of scope here — happy to do it as a follow-up if you'd prefer.

`go build`, `make lint` (0 issues), and the touched-package tests are green; fail-hard paths re-verified via the CLI.

dpup:
Added the regression test you suggested: `TestRunProviderPreflightUsesResolvedConfig` in `internal/cli/provider_test.go`. It sets up a repo + a worktree branch whose `moat.yaml` diverges (`name: from-worktree` vs the main workspace's `from-main`), runs `RunProvider` with `cli.DryRun=true` via a real `testprov` command, and asserts `Preflight` observes the worktree's config (plus the no-worktree companion). I verified it **fails** against the pre-fix ordering (sees `from-main`) and passes with the fix — so it'll catch any future config-source drift. `make lint` clean, tests green.

PR #434, opened 2026-07-03
feat(pi): build-time Pi packages + Moat-owned safe settings

dpup:
## Summary

Bridges Pi's configuration into Moat with two coupled, **build-time** features:

- **`pi.packages`** — declare Pi packages (extensions/skills/prompt-templates/themes) as remote sources (`npm:` / `git:` / `https://` / `ssh://`); Moat runs `pi install` for each at **image build time**, baking them into a reproducible, cached layer.
- **Moat-owned safe defaults** — every `moat pi` image bakes a `~/.pi/agent/settings.json` a workspace cannot override: `defaultProjectTrust: never` (a checked-out repo's own `.pi/` extensions — arbitrary code — do not auto-load), telemetry off, `quietStartup`.
- **Network warning** — because Pi config (`baseUrl`/custom providers/extensions) can redirect model traffic to any host, only the network policy actually contains egress; `moat pi` now warns under a permissive policy.

## Why build-time (not runtime staging)

`pi install` and the security settings write the *same* file (`~/.pi/agent/settings.json`). Runtime staging would clobber the packages array `pi install` wrote. One build step owns the file — no `moat-init.sh` change, no conflict, and it fits Moat's reproducible-image model. `httpProxy` is intentionally **not** baked (Moat already injects `HTTP_PROXY`).

## How it works

Mirrors the claude-plugins Dockerfile-snippet pattern: `pi.GenerateDockerfileSnippet` emits a script (`COPY --chown` + `RUN` as `moatuser`) that installs the packages then merges the safe settings via `node`. Wired through `ImageSpec.PiBakeSettings/PiPackages` → `GenerateDockerfile` → `ImageTag`. The tag **content-hashes the generated bake script**, so changing the baked defaults or the package set invalidates cached images automatically.

## Verification

- **End-to-end (DIND):** `moat pi` with `pi.packages: ["npm:chalk"]` builds; `pi install` runs at build; the baked `~/.pi/agent/settings.json` is `{packages:["npm:chalk"], defaultProjectTrust:"never", enableInstallTelemetry:false, enableAnalytics:false, quietStartup:true}`; `pi list` inside the image shows `chalk`.
- Network warning verified both ways (shown under permissive, silent under strict).
- `go build`, `make lint` (0 issues), unit tests green. Self-reviewed with an adversarial multi-agent pass; the confirmed findings (content-hash cache key, bare-scheme validation) are fixed in this branch.

## Scope / deferred

Remote sources only (local paths rejected — they record relative paths that break at runtime). A package's runtime deps must be declared in `dependencies:`. Workspace `.pi/` trust opt-in, `pi.settings:` passthrough, `pi.mcp`, and `pi.sync_logs` remain future work.

Design + plan: `docs/plans/2026-07-03-pi-config-bridge-design.md`, `...-plan.md`.


dpup:
Thanks — addressed the nit: added a `ui.Warn` in `manager_create.go` when `pi.packages` is set but `pi-cli` isn't an installable dependency (the `moat run --agent pi` without `pi-cli` case), so it fails loud instead of silently skipping the packages. `moat pi` is unaffected (it always adds `pi-cli`). Build + lint + tests green.

dpup:
Re the cosmetic nit on `cli.go:118` — the `(empty == permissive)` comment is intentional and load-bearing: when there's no `moat.yaml`, `config.Load` returns nil and `RunProvider` substitutes an empty `&config.Config{}` (so `Network.Policy == ""`) *before* `ConfigureAgent` runs — the normalization in `config.Load` doesn't apply on that path. So `!= "strict"` correctly warns for a no-config `moat pi` run. Leaving the comment as-is since it documents exactly that case.

PR #435, opened 2026-07-07
feat: support podman via its Docker-compatible socket

grinnellian:
## Why

Podman is the packaged default container engine on a growing share of Linux distributions (Fedora/RHEL family) and a common choice where Docker Desktop's licensing or its daemon model is a blocker. Supporting it removes a runtime prerequisite for those users — anyone who can run podman gets moat's isolation, credential injection, and observability with no Docker install, and daemonless/rootless setups become first-class. Since podman already speaks the Docker API, moat gets this at near-zero maintenance cost: no new runtime implementation to carry.

## Summary

Podman's Docker-compatible API works with moat's Docker runtime unmodified (verified against podman 6.0.0 / compat API v1.44, macOS arm64), so this PR adds podman as a supported runtime **without a new `Runtime` implementation** — podman is the existing `DockerRuntime` pointed at a podman socket:

- **Auto-detection** — podman sockets join the existing alternative-socket probe (Rancher Desktop precedent): podman machine API sockets on macOS (`$TMPDIR/podman/*-api.sock`), rootless (`$XDG_RUNTIME_DIR/podman/podman.sock`) and rootful (`/run/podman/podman.sock`) on Linux.
- **`--runtime podman`** / `MOAT_RUNTIME=podman` / `runtime: podman` — forces the docker runtime over a podman socket; verifies an explicit `DOCKER_HOST` actually answers as podman (`/version` Components contains "Podman Engine"); errors with start hints otherwise.
- **`moat doctor`** — labels the engine (`Available: docker (podman), …`), treats unknown engine identity honestly (no fail-open), points at an idle podman socket when Docker is unreachable, and no longer trusts podman's gVisor report: podman's compat `/info` lists every containers.conf OCI runtime (`runsc` included) whether or not it's installed.
- **Per-run engine endpoint persistence** — run metadata now records the `DOCKER_HOST` endpoint (`docker_host`, additive field) and lifecycle commands reconnect to it via a host-pinned pool entry. Without this, a run created on a podman (or Rancher Desktop) socket while a real Docker daemon is also reachable would reconnect to the wrong engine on `moat stop`/`logs` — container-not-found plus a leaked container. Fixes that pre-existing hole for Rancher Desktop too.
- **Forced-docker hardening** — an explicit `--runtime docker` no longer silently falls back to a podman socket (the mirror image of the podman path's engine verification).
- **Docs** — installation (incl. the podman 6 libkrun/krunkit provider pitfall on macOS), runtimes concept, CLI/env/moat.yaml references, troubleshooting, comparison table, README, CHANGELOG.

## Verification (all on a live podman machine)

- `moat run --runtime podman -- <cmd>` with no `DOCKER_HOST`: probes, finds the machine socket, pulls, builds the derived image (all 9 steps), runs, exits 0.
- Credential injection hard floor: `moat run --runtime podman --grant github` — container env contains only the placeholder (`GH_TOKEN=ghp_moatProxyInjectedPlaceholder…`); `gh api user` inside the container succeeds via proxy header injection. Per-run token proxy auth and `host-gateway` resolution work under podman machine (requires podman ≥ 4.1).
- Wrong-engine trap: run created via podman while a real Docker daemon is reachable → fresh-process `moat list` shows it running and `moat stop` stops the podman container (previously reconnected to Docker).
- `go build`, `go vet`, `make test-unit` (race) clean; new tests use a fake Docker-API `httptest` server — no live daemon needed in CI.

## Notes

- gVisor under podman on Linux: indistinguishable via the compat API whether `runsc` is actually installed, so sandbox behavior is unchanged; moat warns once that the report is unverified, and doctor + troubleshooting docs explain the failure mode and `--no-sandbox`.
- The CHANGELOG entry's `#NNN` placeholder will be filled with this PR's number as a follow-up push.


grinnellian:
Closing temporarily pending an internal review of the fork — will reopen shortly.

grinnellian:
For transparency: this PR (authored by an AI agent — Claude, working from a mission brief) was pulled back by the human operator, who wants to review the fork themselves before it's submitted for upstream review. It will be reopened once that review is done. The temporary-close comment above was also posted by the agent; pulling it back was the human's call.

PR #437, opened 2026-07-09
feat(copilot): add context and reasoning_effort config fields

abezzub-dr:
Add `copilot.context` and `copilot.reasoning_effort` to moat.yaml, with corresponding `--context` and `--reasoning-effort` CLI flags. CLI flags override config values, matching the existing `--model` precedence.

This lets users pin context window tier (`default`/`long_context`) and reasoning effort level (`none`/`minimal`/`low`/`medium`/`high`/`xhigh`/`max`) declaratively so settings persist across moat runs.

### moat.yaml example

```yaml
copilot:
  model: claude-opus-4.6
  context: long_context
  reasoning_effort: high
```

### CLI usage

```bash
moat copilot --context long_context --reasoning-effort high
```

PR #438, opened 2026-07-09
feat(copilot): carry over host settings into container

abezzub-dr:
Read `~/.copilot/settings.json` and optional `~/.moat/copilot/settings.json` overrides, filter through an allowlist of safe/useful settings, and write the merged result to the container staging directory.

## Settings carried over

contextTier, effortLevel, footer, includeCoAuthoredBy, model, mouse, subagents, tabs, theme

## Security: statusLine

`statusLine.command` executes arbitrary commands inside the container. It is **only** allowed from `~/.moat/copilot/settings.json` (explicit user opt-in), not from the host's `~/.copilot/settings.json`.

## Precedence

CLI flag / moat.yaml > settings.json > Copilot default

Model/effortLevel/contextTier are stripped from settings.json when the corresponding CLI flag or moat.yaml field is set.

## Design

- `MergeSettings(MergeOpts{})` accepts override values as parameters (no global state coupling)
- `MOAT_SKIP_HOST_COPILOT_SETTINGS=1` disables the feature for tests/CI
- Follows the same pattern as Claude Code's settings merge (RawExtras passthrough)


abezzub-dr:
this has a dependency on https://github.com/majorcontext/moat/pull/437, so I will rebase + update this PR once that lands

PR #442, opened 2026-07-21
feat(deps): add gitleaks secret scanner dependency

abezzub-dr:
## What

Adds [gitleaks](https://github.com/gitleaks/gitleaks) as a first-class `github-binary` dependency, so it can be declared in `moat.yaml`:

```yaml
dependencies:
  - gitleaks
```

Previously `dependencies: [gitleaks]` failed with `unknown dependency "gitleaks"` — the field only accepts registry entries and the dynamic prefixes (`npm:`, `pip:`, `go:`, …). The only workaround was `go:github.com/gitleaks/gitleaks/v8`, which drags in the Go toolchain and compiles from source.

## Why

gitleaks is the de-facto standard secret scanner and the tool behind the common pre-commit hook. Agents running a repo's pre-commit hooks in a moat container need the binary present.

## How

- New `github-binary` entry in `internal/deps/registry.yaml`, modeled on `gh`/`opentofu`. Installs from the prebuilt release tarball — no Go toolchain, one image layer.
- The amd64 release names its arch `x64` (not `amd64`), so the `targets` map remaps it.
- Binary name matches the dependency name, so no `command:` override.

## Tests

- `TestRegistryHasGitleaks` — asserts the spec shape, the `x64` amd64 remap, tar.gz extraction, install path, and asset URL for both the default and an explicit non-default version (companion case).
- Covered by the existing `TestRegistryGithubBinaryPlaceholders` (no unsubstituted placeholders) and `TestRegistryGithubBinaryURLsExist` (live HEAD check of the real download URLs — passes).

## Docs

- Added to the Development tools row in `docs/content/reference/06-dependencies.md`.
- CHANGELOG entry under Unreleased → Added.


abezzub-dr:
@dpup any reason the dependency registry is a hardcoded list in the repo? It seems nice to allow project's moat.yaml to declare any dependency it wants and have moat install it during image build

dpup:
The idea was the only hard coded dependencies either needed custom code/orchestration, or were super common syntactic sugar.

For moat.yaml would the hooks suffice?

dpup:
What about adding `github:` as a common path for binaries?

abezzub-dr:
hooks worked for this dependency 👍

Issue #396, opened 2026-06-23
Add OS-native kernel sandboxing (Landlock/Seatbelt): in-container defense-in-depth + lightweight local mode


dpup:
## Summary

Add OS-native kernel sandboxing (Landlock on Linux, Seatbelt on macOS) as an enforcement layer for Moat runs. Support it in **two modes**:

1. **In-container defense-in-depth** — apply kernel self-restriction to the agent process *inside* a Moat container, so a container-boundary breakout still hits an irreversible, deny-by-default capability wall.
2. **Lightweight local mode** — a containerless run mode that uses kernel primitives alone (no Docker/Apple container), for fast, low-friction local runs where full container isolation is more than needed.

## Motivation

Moat's current isolation model is the container boundary plus the credential-injecting proxy. That's strong, but:

- **Single layer of process isolation.** If an agent escapes the container (kernel/container-runtime CVE, misconfig), there's no inner wall. Kernel self-restriction applied to the agent process gives defense-in-depth: even with arbitrary code execution, the agent cannot widen its own restrictions.
- **Startup cost / footprint.** Spinning up a container (image selection, mounts, daemon registration) is overhead for quick local runs. A kernel-only mode could start in well under a second with no disk usage, lowering the barrier to "just sandbox this one command."
- **Defense-in-depth aligns with Moat's deny-by-default posture.** A curated default blocklist of sensitive paths (SSH keys, `~/.aws`, `~/.config/gcloud`, shell history, browser cookies) enforced at the kernel level complements the proxy's credential isolation.

## How it would work

**Linux — Landlock LSM (kernel 5.13+):**
- Agent process calls `landlock_restrict_self()` after setup, before exec'ing the agent.
- Restrictions are irreversible (tighten-only) and inherited by all child processes — no userspace escape hatch.
- Allow-list filesystem access (typically workspace CWD writable, everything else denied) and, on kernel 6.7+ (ABI v4), TCP network rules.

**macOS — Seatbelt:**
- Apply a sandbox profile to the agent process for filesystem/network restriction.

**Mode 1 (in-container):** the container entrypoint applies kernel restrictions to the agent process before handing off. Container boundary = outer wall; Landlock/Seatbelt = inner wall.

**Mode 2 (local):** `moat run` skips container creation entirely; the agent runs as a restricted host process. The credential proxy still sits outside the sandbox (localhost), so the existing per-run token injection model carries over unchanged.

## Configuration sketch

```yaml
# moat.yaml
isolation:
[...]

PR #443, opened 2026-07-23
feat(sandbox): Landlock kernel sandbox for the agent process (in-container)

dpup:
First cut of #396: OS-native kernel sandboxing, in-container mode.

## What this does

`isolation.kernel_sandbox: true` in `moat.yaml` (or `moat run --kernel-sandbox`) applies a [Landlock](https://docs.kernel.org/userspace-api/landlock.html) filesystem sandbox to the agent process inside the container — an inner wall behind the container boundary. The restriction is applied after the entrypoint's privilege drop, is inherited by every child process, and cannot be widened once applied: even code with arbitrary execution inside the run stays behind it.

The policy is a write allowlist (reads work everywhere):

| Writable | Everything else |
|---|---|
| `/workspace`, agent `$HOME`, `/tmp`, `/var/tmp`, `/dev` (+ioctl), `/proc`, `/run`, read-write mount targets, `isolation.sandbox.allow_write` entries | write-denied by the kernel, regardless of file permissions |

```yaml
isolation:
  kernel_sandbox: true
  sandbox:
    allow_write: [/data]   # optional extras
```

## How it works

- **`internal/sandbox`** — policy computed host-side from the final mount set (pure, unit-tested), JSON-shipped via `MOAT_SANDBOX_POLICY`, applied in-container with [go-landlock](https://github.com/landlock-lsm/go-landlock) v0.9.0 (pure Go, by the Landlock maintainer). Best-effort ABI downgrade; a probe distinguishes "downgraded" from "unavailable" so degradation is always announced.
- **`cmd/moat-sandbox`** — tiny static helper that self-restricts and execs the agent. Restrict-then-exec on one goroutine sidesteps go-landlock's multi-thread caveat. Fails closed if the policy is missing; logs `kernel sandbox active (Landlock ABI vN)` (or a warning) into the run's logs.
- **`internal/sandboxbin`** — ships the helper into run images by embedding prebuilt linux/amd64+arm64 blobs in the moat binary, deliberately mirroring the `internal/initbin` pattern from #441 (committed fail-closed stubs, `go generate` wired into `make build-cli`/goreleaser, checksums + commit-guard tests) so the two merge cleanly — once #441 lands, this can fold into the Go moat-init.
- **`moat-init.sh`** — final exec becomes `exec [gosu moatuser] moat-sandbox "$@"` when the policy env is set; fails closed if the helper is missing from the image.
- **Image pipeline** — `ImageSpec.NeedsKernelSandbox` forces the init entrypoint + custom image and content-hashes the helper into the image tag, so toggling the sandbox (or a rebuilt helper) rebuilds the image. The run manager fail-fasts with rebuild instructions when the embedded helper is a stub (`go install` builds) or the arch is unsupported.

## Design decisions (documented in `docs/plans/2026-07-23-kernel-sandbox-design.md`)

- **Write-allowlist posture** (reads open) matches Codex CLI's Landlock policy and Anthropic sandbox-runtime's `allowWrite` model; domain-level read-secrecy stays with the credential proxy.
- **`deny_paths` is rejected, not ignored** — Landlock is allowlist-only; a user who wrote `deny_paths` and got no error would believe those paths are protected. Same for `isolation.mode: local` (the containerless mode is a #396 follow-up).
- **Best-effort with loud degradation** — pre-5.13 kernels, Docker <23 seccomp, and gVisor lack Landlock; the run proceeds with a warning in `moat logs` (per the issue's "warn + degrade" requirement). Honest constraint: Landlock's `no_new_privs` disables `sudo` inside sandboxed runs; `pre_run` hooks run before the sandbox and are unaffected.
- **Network rules deferred** — Landlock v4 TCP rules interact with the proxy routing open questions in #396; filesystem-only for the first cut.

## Testing

- Unit (race detector): config validation both directions, policy construction (rw-mount inclusion **and** exclusion contracts), JSON round-trip + garbage rejection, Dockerfile/context-file generation with companion absence case, image-tag divergence, stub/checksum guards.
- **Real enforcement test** (`internal/sandbox/apply_linux_test.go`): a sacrificial subprocess self-restricts and asserts allowed writes pass, denied writes fail, reads stay open (skips on non-Landlock kernels).
- **E2E** (`internal/e2e/kernel_sandbox_test.go`): full docker run with a build-time `moatuser`-owned probe dir outside the allowlist — write denied *only* under the sandbox (companion run proves it's Landlock, not DAC), `allow_write` honored, `kernel sandbox active` line asserted. Both pass against real Docker (Landlock ABI v6).
- Manual CLI validation: `moat run --kernel-sandbox . -- sh -c ...` → `kernel sandbox active (Landlock ABI v6)`, `/workspace` ok, `/etc` denied, reads ok.
- `golangci-lint run`: 0 issues. Full unit suite green except the pre-existing network-dependent `TestRegistryGithubBinaryURLsExist` (fails on main in this environment too).

## Follow-ups tracked in #396

macOS Seatbelt, containerless `isolation.mode: local`, `deny_paths` (needs masking mounts), Landlock TCP rules scoped to the proxy, audit-store surfacing of kernel denials.


dpup:
Closing after a threat-model + runtime-coverage review (full write-up going to #396).

Short version: the in-container Landlock sandbox is a no-op on both of moat's default runtimes — Linux defaults to gVisor (which doesn't implement the landlock syscalls), and Apple containers ship a minimal kernel with the entire LSM framework compiled out (`CONFIG_SECURITY=n`; confirmed empirically — `/proc/sys/kernel/landlock/version` is absent inside an Apple container). It only enforces on `--no-sandbox` Linux runs and macOS Docker (LinuxKit), which is too thin a slice to justify a permanent security surface with a 'does nothing on your setup unless…' asterisk.

The work isn't lost: the branch, design doc, the cross-platform policy abstraction, and the binary-embedding pattern are the natural starting point for kernel sandboxing in a future containerless/local execution mode, where the agent is a native process and Landlock (Linux) / Seatbelt (macOS) actually become the primary wall. Keeping the branch `feat/kernel-sandbox` for that.

Detailed findings in #396.

PR #444, opened 2026-07-25
feat(container): support podman via its Docker-compatible socket

grinnellian:
This reopens the work from #435, opened by an over-eager Claude. Now deeply reviewed by human eyes; sorry about the prior.

Four questions first, because these are the parts I don't think I should decide on my own.

**1. gVisor under podman — is a warning enough?**

Podman's compat API reports every OCI runtime listed in `containers.conf` as available whether or not the binary is installed, so a reported `runsc` proves nothing. That means moat cannot verify the isolation it promises when it runs on podman. This PR warns once, names the command to check, and turns a creation failure under `runsc` into a targeted diagnostic instead of a raw OCI error — but it still proceeds. For a tool whose whole pitch is the sandbox, that may be the wrong default.

I can see three answers: require an explicit acknowledgement for podman + sandbox, verify `runsc` out of band, or leave it and document the caveat. I didn't want to pick unilaterally, and I'd rather drop podman from the supported matrix than have it quietly weaken the guarantee. Happy to implement whichever you prefer.

**2. Should `podman` exist as a runtime value at all?**

It isn't a separate `Runtime` implementation but rather`DockerRuntime` pointed at a podman socket, and a run still records `Runtime: "docker"`. There's value in asserting engine identity: a mis-set `DOCKER_HOST` would otherwise run on Docker silently. But if you'd rather this were just `DOCKER_HOST` plus documentation, that's a much smaller change and I'd understand the call.

**3. Would you rather this were split?**

It's roughly 1,200 lines across three separable pieces: per-run engine pinning (which fixes Rancher Desktop independently of podman), podman detection and labeling, and the `moat stop` behavior change. I kept it as one PR because the first piece doesn't fully justify itself without the second but they should separately reviewable, and I'll split it on request.

**4. Is the `moat stop` change in scope here?**

It changes behavior for non-podman users too. Previously a not-found container only warned and still recorded the run stopped, which can mark a run stopped while its container keeps running on another engine. It now fails loudly — but only when a *different, live* podman endpoint is reachable, so a container that is simply gone still stops cleanly. If you'd rather that were its own PR, or an issue first, I'll pull it out.

## What this does

Podman's Docker-compatible API works with moat's existing Docker runtime unmodified, so this adds podman without a new `Runtime` implementation:

- **Auto-detection** — podman machine sockets (macOS) and rootless/rootful sockets (Linux) are probed exactly like Rancher Desktop's, only when the default Docker socket is unreachable and `DOCKER_HOST` is unset.
- **Explicit selection** — `--runtime podman`, `MOAT_RUNTIME=podman`, or `runtime: podman`. Forcing podman errors if a usable podman engine can't be reached, so it can't silently succeed against real Docker.
- **Engine identity is recorded, not guessed** — asked of the daemon at run creation and persisted, so `moat list` / `moat status` label the engine without inferring it from a socket path.
- **Per-run endpoint pinning** — each run records the endpoint it was created on, so lifecycle commands reconnect to the right engine when several are present.

## What the self-review changed

The things worth calling out, since they were wrong before:

- **The version floor was wrong.** The branch claimed podman 4.1+. `--add-host host-gateway` landed in podman **4.7.0**, and the requirement is Linux-only — `synthHostStrategy` emits that sentinel only for a Docker-API runtime on a Linux host, so the macOS rows were citing a constraint that never applied to them.
- **Three doc claims contradicted the code**, including one telling users to recover with `moat destroy --force` after that teardown had been given its own flag.
- **The `moat stop` guard was too broad** — it fired whenever a podman socket *file* existed, so `docker system prune` on a podman-equipped host broke `moat stop`. It now requires a reachable, distinct engine, compared by filesystem identity rather than by path string (with `podman-docker` installed, `/run/docker.sock` is a symlink to the podman socket, so string comparison sees two engines where one exists).
- **Engine identity was inferred from the socket path**; it's now recorded authoritatively at creation.
- Test seams were consolidated out of the shipped binary, the pool's negative cache got a TTL, and an identity probe that ran on every startup was removed.

## Known gaps

- **e2e coverage for the `destroy` running-run guard is unit-level only.** I didn't add an e2e case I couldn't run.
- **Windows isn't auto-detected** — there are no podman socket candidates for it. A manual `DOCKER_HOST` isn't blocked by anything in moat, but I haven't verified it end to end and haven't claimed it works.
- **Version floors are external facts.** moat enforces no version gate in code; the table reflects podman's release notes and what I actually exercised (6.0.0 / 6.0.1, macOS arm64), not a matrix I've tested.

## Notes

- The CI e2e job logs into Docker Hub with repository secrets, which GitHub doesn't pass to fork PRs — so it will likely fail at that step here, before any test runs. Not related to these changes, but I'd rather flag it than have it look like a red check I ignored.
- The two CHANGELOG entries currently link to #435; I'll point them at this PR's number once it has one.
- `internal/providers/claude.TestResolveClaudeCredential_SkipsMigrationInReadOnlyDir` fails in any root container (root bypasses the `0500` permission it relies on). It fails identically on `main` — flagging so it isn't mistaken for fallout from this branch.

Verified on the branch: unit suite green under `-race`, `go vet` clean including under the `e2e` build tag, and `golangci-lint` at exactly the same nine pre-existing `prealloc` findings as `main` — no new lint.

As with #435, and to keep that disclosure attached to this one: this branch was written by an AI agent (Claude) working for me, and I've reviewed (and actually run) it before submitting. Happy to answer for any of it.


dpup:
Hey @grinnellian, thanks for the PR. It would be great if you split this up, it's of a bear to review.

grinnellian:
Super fair @dpup! It does not decompose easily into useful parts, but i will take another pass at trying to get it bite-size. Thanks for taking a look at all at a massive unrequested feature! Podman support is already super useful in my home lab so i wanted to share it back to the community. Thanks so much for working on `moat`; it has been an amazing accelerator for me.

dpup:
Thanks, appreciate it.

For the followup, there's some interface level changes that could be pulled out. In terms of podman should gVisor fail hard since it's unproven?

grinnellian:
Copy that, thanks for the breadcrumb!
gVisor: i was wondering this too & will have it fail hard in the follow-ups

PR #446, opened 2026-08-01
docs(release): cut v0.7.0 changelog and sync doc indexes

dpup:
Release prep for **v0.7.0**. Docs-only — no Go changes.

## CHANGELOG

`## Unreleased` → `## v0.7.0 — 2026-08-01`, and the summary paragraph rewritten. The old summary led with Keep body policies and volumes and never mentioned `moat pi` or `moat copilot` — the two headline additions of this release.

Audited every commit since `v0.6.1` (39 of them) against the entry list: all user-facing changes are already recorded. The rest are refactors (`#420`–`#429`), lint/format/CI (`#413`–`#419`, `#432`), and follow-up polish folded into an existing entry (`#411` into the port-in-use entry).

## Docs audit

Checked each unreleased feature for reference + guide coverage — `moat open`, the discovery index, `opentofu`/`terragrunt`, `type: volume`, `workspace.mode`, `params.body`, `MOAT_KEYRING_BACKEND`, `moat copilot`, `moat pi`, `pi.packages`, `reasoning_effort`. **All present.** The `docs/content` pages are in sync; the gaps were in the index/overview surfaces around them:

- **`docs/README.md`** was missing nine pages — the worktrees, workspace-sharing, volume-workspaces, multi-agent, recipes, Pi, and Copilot guides, plus the provider-yaml and troubleshooting references. Several predate this release.
- **`README.md`** listed `moat attach`, which no longer exists (verified against `moat --help`), and omitted `moat gemini`, `moat pi`, and `moat open`. Added Gemini and Pi quick-start sections alongside the existing three agents.
- **`docs/content/reference/04-grants.md`** documented which agent uses each credential for Claude, Codex, Gemini, and Copilot, but never mentioned Pi. Added a pointer from both the Anthropic and OpenAI sections to the Pi guide's backend-resolution rules.

No version pins exist in `docs/content`, and the goreleaser build takes its version from the tag, so nothing else needs bumping.

## Verification

- Every relative link added or touched resolves to a real file (checked).
- No `#NNN` placeholders remain in the CHANGELOG (CI gate).
- `moat grant gemini` and `moat grant providers` output verified against a live build before documenting them.

## Not included

Two things surfaced during the audit that are code, not docs, so they are not in this PR — see the review comment below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@ Moat runs AI coding agents in isolated containers with credential injection, net
 
 Moat is pre-1.0. The CLI interface and `moat.yaml` schema may change between minor versions. Breaking changes are listed under **Breaking** headings below.
 
-## Unreleased
+## v0.7.0 — 2026-08-01
 
-Adds HTTP request-body inspection to Keep policies. File- and pack-based `network.keep_policy` rules can now match on the parsed JSON request body, so policies can enforce content-based rules (e.g. block requests whose body carries a secret) instead of host/method/path alone. Also adds native volume backing for `volumes:` (`type: volume`) — a Docker named volume on the engine's native filesystem, for container-only working directories that should bypass the host↔VM filesystem-sharing layer a bind mount crosses. The routing proxy now serves a discovery index at its bare hosts so you can browse an agent's endpoints instead of memorizing hostnames.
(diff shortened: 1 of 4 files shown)


dpup:
Two code-level findings from the docs audit, left out of this PR to keep it docs-only.

**1. `moat grant providers` output is misleading.** Live output on this branch:

```
anthropic      Anthropic API key or OAuth credentials   builtin
anthropic                                               builtin
...
meta                                                    builtin
oauth                                                   builtin
pi                                                      builtin
```

- **Duplicate `anthropic`, one blank.** `claude` (the OAuth provider) is displayed under its CLI alias `anthropic` via `goProviderCLINames`, while `AnthropicProvider` also registers under the real name `anthropic` (`internal/providers/claude/provider.go:30-31`). Both rows render; only the aliased one has a description.
- **`pi` is listed but is not grantable.** `internal/providers/pi/provider.go:34` — `Grant()` always returns "pi has no credential of its own". `copilot`, the identical agent-only case, is explicitly skipped at `cmd/moat/cli/grant_providers.go:59-61`; `pi` was never added to that skip.
- **Blank descriptions** for `meta`, `oauth`, and `pi` — they are absent from `goProviderDescriptions` and don't implement `DescribableProvider`.

**2. The apt Python fallback ignores the requested version** (pre-existing, unrelated to this release). With two runtimes the base is `debian:bookworm-slim` and `python@3.12` silently installs whatever `python3` Debian ships — 3.11 on bookworm. It should honor the version or warn. This is the same code path as #445.

Neither blocks the v0.7.0 tag. Happy to fix (1) in a small PR before tagging if you want the table clean for the release — it's a few lines in `grant_providers.go` plus a test.

PR #450, opened 2026-08-03
fix(tui): stop moat overriding terminal state that inline agents drive

dpup:
Three fixes with one root: moat sits between the agent and the terminal, and was asserting authority over state the agent legitimately drives. Claude Code never exposed any of it — it uses the alternate screen and leaves the keyboard protocol alone. Codex renders inline and negotiates with the terminal directly, so it hit all three, which together made it hard to use inside moat.

Every fix below was diagnosed from bytes captured with `--tty-trace`, not from reading the spec. In two places the spec reading was wrong.

## 1. `ctrl+/` did nothing under Codex

moat's escape prefix matched a single byte, `0x1f` — the legacy encoding of a chord that is ambiguous with `ctrl+_`. Codex pushes `CSI > 7 u` at startup to enable the kitty keyboard protocol, moat forwards that to the terminal, and the terminal then reports `ctrl+/` as `CSI 47;5u`. Nothing matched, so **every** command behind the prefix was unreachable under Codex — including `ctrl+/ d`, the TUI dump that exists to debug exactly this.

The matcher now understands the CSI-u form alongside the raw byte, and is narrow: only key `47` with the Ctrl modifier, so other kitty-encoded keys reach the agent untouched. Captured bytes showed `ctrl+c` and `ctrl+d` are CSI-u encoded too, and there is a test asserting they pass through — a sloppier matcher would have eaten `ctrl+c` and made Codex uninterruptible.

The real bug took four attempts to find, because three of them fixed the matcher when the matcher was never in the failing path. When the read buffer ended while the prefix was armed, `Read` took a **one-byte shortcut** to grab the command key; terminals deliver the chord's release in its own read, so that byte was the `ESC` beginning the release sequence. The trace showed moat emitting:

```
\x1b[?62;22;52c   \x1f   \x1b[47;5:3u   d
```

— a literal prefix, the leaked release, and the command key handed to the agent. That path now defers to the sequence matcher.

Key-up and autorepeat events are also swallowed while the prefix is armed; neither can be a command key, and treating either as one cancelled the prefix mid-chord.

## 2. Prompt area smeared, cursor misplaced, content shifting wholesale

moat replaced **every** scroll region the child requested with its own, so the footer row could not be scrolled away (#349, where Ink's defensive `ESC[r` let footers trail into scrollback).

That assumes the child never needs a region of its own. Codex uses scroll regions as a drawing primitive: restrict to a band, reverse-index (`ESC M`) to open space inside it, reset, redraw. With the band replaced by the full content area, those reverse-indexes scrolled the entire screen on every update.

moat now clamps rather than discards — the child's band is forwarded with its bottom margin held above the footer row. moat still owns that row, so #349's guarantee holds, and a bare `ESC[r` (that PR's actual case) still becomes moat's region. All twenty regions Codex requested in a captured session sit inside the content area and now pass through unaltered; that set is pinned as a test.

## 3. Footer vanished until the agent settled

`CSI J` — "erase from the cursor to the end of the display" — reaches past the child's content area into the footer row. moat repaired the footer only on a redraw debounced behind 50ms of quiet, which never arrives during continuous rendering. The capture showed five of them, issued from rows 4, 18 and 51 of a 95-row terminal.

The footer is now repainted immediately after the erase, cursor preserved. Codex wraps frames in synchronized output, so the repaint lands atomically with the frame.

## Also

`MOAT_TTY_TRACE`, an env equivalent of `--tty-trace`. The sessions worth tracing are the broken ones, where `ctrl+/ d` may be unreachable by definition — as it was here.

## Verification

Manually QA'd against both Codex and Claude by @dpup. Unit tests cover each fix, and each was mutation-tested rather than merely run green — reverting any one guarantee fails the suite with the symptom it describes, e.g. dropping the region clamp yields `"\x1b[1;95r" clamped to "\x1b[1;95r", which includes the footer row`.

## Review attention

- **`#349` regression risk.** The bare-reset case is preserved exactly, but `TestWriter_InterceptsDECSTBM_WithArgs` previously asserted that an explicit band *is* swallowed. That test now asserts clamping. If footers start trailing in scrollback under Claude, this is the cause.
- **The inline footer repaint.** RIS had precedent, but the DECSTR path deliberately avoids inline redraws to protect mid-frame rendering. An agent that erases the display without synchronized output could flicker; the fallback would be to mark dirty instead.
- **Not verified:** an early report of odd background colours. Plausibly the scroll-region bug moving styled rows, but never isolated, so treat it as untested rather than fixed.


dpup:
Both bugs confirmed and fixed in 9dbae8b. I reproduced each behaviorally before touching anything.

**1. Bare `Esc` withheld** — real, and worse than what this branch set out to fix. A probe measured the delay:

```
first Read returned 2 bytes "\x1bx" after 82ms
```

The `Esc` only escaped when the *next* key arrived — unbounded in practice, and `Esc` is common in both agents ("esc to interrupt" in Codex). You were right that this is a regression introduced here; before the branch a bare `0x1b` passed straight through.

Fixed as you suggested: a lone `0x1b` is held only while the prefix is armed, where it is most likely the chord's release arriving in its own read. The split-read case that motivated the buffering still works, because a terminal splitting mid-sequence delivers `\x1b[4` rather than a lone `\x1b`.

**2. Partial dropped at EOF** — also real, same reproduction as yours:

```
child received: "\x1f"      // the 0x1b vanished
```

Both EOF paths now flush everything held, in order, rather than returning after the prefix.

Regression tests added for both, and both are mutation-tested: restoring the unconditional hold fails with `Esc was withheld for 65ms awaiting a following key`.

**Minor findings, all taken:**

- `ctrlED` now matches only `Ps=0`/`Ps=2`. Your reasoning holds — `Ps=1` cannot pass the footer because the child's screen ends a row above it, and `Ps=3` only clears scrollback. Both now flow through untouched, with a test.
- Footer repaint bytes extracted to `footerRepaintBytes()`, shared with `redrawFooterLocked`. I deliberately left `ctrlRIS` on its own: it homes the cursor afterwards rather than restoring it, because RIS itself homes the cursor and the child resumes from there. Folding it in would have changed behavior, so the difference is now documented at the helper.
- `--tty-trace` / `MOAT_TTY_TRACE` precedence extracted to `resolveTracePath` and tested in all four combinations.
- Good catch on `.gitignore` — the pattern didn't match the documented filename. The env-var docs now use `session.tty`, and the docs say explicitly that traces are JSON regardless of suffix and why `.tty` is the one that gets ignored.

Thanks — the two input bugs would have been unpleasant to track down in a live session, and the `Esc` one would have hit Claude Code users too, not just Codex.

dpup:
All three fixed in 1491e00, each reproduced first.

**1. Repeated splits** — confirmed with your exact scenario, a release split three ways:

```
actions=[] out="\x1f;\x1b[47d"
```

You identified the root correctly: the one-byte fallback ran even with `partialSeq` non-empty. It is now skipped whenever a partial is held, so the tail path re-enters `Read` and the main scanner resumes with the buffered bytes prepended. Regression test uses the five-chunk split from your probe.

On the "flagged in the previous review" framing — the earlier pass flagged the *first* split (release arriving in its own read), which 8053cd9 fixed. This is the second-order case, and it was genuinely still open. Either way the fix needed to be general rather than one more special case, which is what the earlier round should have done.

**2. Multi-level pop** — confirmed:

```
after CSI<2u, cleanup emitted: "\x1b[<u\x1b[r\x1b[2J\x1b[H\x1b[?25h"
```

`kittyPopCount` now parses `Pn`, defaulting to 1. This one mattered most: the spurious pop stripped a level moat never owned, which is precisely the hazard the feature exists to prevent, so the code contradicted its own comment. Tested across two-level, three-level, partial, and over-large pops.

**3. Split push across writes** — confirmed: `"\x1b[>7"` then `"u"` never counted, so cleanup restored nothing. The incomplete-CSI branch now covers the `>`/`<` lead bytes. Good catch that container stdout chunks arbitrarily; this would have been intermittent and near-impossible to attribute in the field.

**4. EOF flush with a small buffer** — agreed it is unreachable through `io.Copy`, and documented in place with the reason, so a future caller with a small buffer does not silently strand bytes.

Also re-verified end-to-end through a real pty: a child that pushes and dies still leaves exactly one pop on the wire.

Worth noting these three were all in code added by the previous review round's fixes — the escape path has now needed four passes, which is a fair signal that a single-byte state machine was the wrong shape for parsing multi-byte sequences. Refactoring it into a proper incremental parser would be the durable fix; I have kept this branch to targeted changes rather than restructuring under review, but it is worth an issue.

PR #453, opened 2026-08-07
fix(claude): stop the anthropic grant overriding Claude Code's OAuth login

dpup:
## The bug

Granting `anthropic` alongside `claude` exported `ANTHROPIC_API_KEY` into the container environment, where Claude Code prefers it over the OAuth credentials Moat had just staged. The session silently billed API credits instead of the subscription and lost OAuth-only capabilities such as the 1M context window. Claude Code reports this only as a passing notice:

> ⚠ claude.ai connectors are disabled because ANTHROPIC_API_KEY or another auth source is set and takes precedence over your claude.ai login

Two code paths disagreed. `containerEnvForCredential` (`agent.go`) already omits the variable for OAuth credentials, with a comment explaining why. The grant loop in `manager_create.go` called `ContainerEnv` for every grant and appended it unconditionally.

## The fix

When both grants are active, the key is staged into a `BASH_ENV` file instead of the container environment.

The mechanism rests on an asymmetry: Claude Code ships as a native binary and ignores `BASH_ENV`, while non-interactive bash — exactly how its Bash tool runs commands (`bash -c '...'`) — sources it on every invocation. Shell commands get a real API key; Claude Code never sees one.

A `claude()` guard keeps nested invocations on OAuth. It unsets `BASH_ENV` as well as the key: a shell-script launcher would otherwise re-source the file and re-export it, silently undoing the unset.

The staging decision is hoisted into a single `willStageClaude` condition shared by the runtime context and the staging block. The context is rendered first and passed *into* staging, so without one condition the generated `CLAUDE.md` could describe a `BASH_ENV` file that was never written. If staging does not run at all (a pi run), the key falls back to the container environment rather than being dropped.

Granting `anthropic` on its own is unchanged — the key is exported normally and Claude Code uses it.

## The injected CLAUDE.md note

The generated context gains a `## Calling the Anthropic API` section. This is load-bearing, not documentation. An agent that notices the placeholder value, or hits a 429 from using the wrong credential, will otherwise "fix" things by exporting the key globally — reintroducing the bug — or go looking for a way to make the OAuth token work for arbitrary calls. The section states that a real key exists, that the placeholder is intentional, to send `x-api-key` explicitly, and not to export it globally.

## Testing

Companion cases throughout, per the repo invariants: staged/not-staged, section present/absent, build flag set/unset.

Four tests exercise the real bash semantics the design depends on, since a silent regression there would either strand the key or leak it to Claude Code:

- key is exported to a non-interactive bash via `BASH_ENV`
- guard strips it from a nested non-bash `claude`
- guard still holds one level down, inside a nested bash script
- ordinary commands keep the key

The guard tests were mutation-checked — with the guard removed the nested process sees `moat-proxy-injected`, so they fail as intended rather than passing vacuously.

`make lint` clean. Unrelated pre-existing failure in `internal/deps`: `TestRegistryGithubBinaryURLsExist/air` gets a 401 on the `cosmtrek/air` release URL (the repo appears to have moved to `air-verse/air`). It fails identically on `main`.

## Limits

- `sh`/dash ignores `BASH_ENV`, so a command launched via `sh -c` will not see the key.
- The guard is a convenience, not a boundary — invoking `claude` by absolute path bypasses it. It prevents an accidental billing switch, not an adversary.
- `BASH_ENV` applies to every non-interactive bash in the container. Two exports and a function is low blast radius, but it is not zero.


dpup:
Both findings were real. Fixed in f947fe5.

## 1. `BASH_ENV` unreadable outside the staging mount's owner — confirmed, fixed

The analysis was right, and I'd missed that my file was the *only* one read from the mount at runtime. Every other staged file is copied out by `moat-init.sh` while it's still root (lines 150-190) and chowned, so the `0700` never mattered for them.

I checked the container I was working in and the mount showed `drwx------ 7 5000 5000` — readable, because Docker Desktop remaps bind-mount ownership. That's what made this invisible locally, and it's exactly why the finding matters: the platforms where it breaks aren't the one in front of me. Combined with bash silently ignoring an unreadable `BASH_ENV`, it fails closed and quiet.

Rather than loosening the staging dir to `0711`, I followed the pattern the other files already use: `moat-init.sh` copies `anthropic-env.sh` into `$HOME/.claude/` as root, and the existing `chown -R "$TARGET_HOME/.claude"` covers it. That sidesteps the mount-ownership question entirely instead of trading it for a weaker directory mode, and keeps `.credentials.json` behind `0700` as-is.

`BASH_ENV` is now `$HOME/.claude/anthropic-env.sh` with `$HOME` left unexpanded — bash expands the value before using it as a path (verified), so it resolves correctly whether `moat-init.sh` targeted `/home/moatuser` or an alternate `$HOME`.

Since nothing structurally ties the Go constant to the shell script and the failure mode is silent, there's now a drift guard in `internal/deps` asserting both the filename and the copy target. It lives there rather than in the claude package because `deps` already imports `claude` and the reverse direction is an import cycle.

## 2. Section missing when there's no `moat.yaml` — confirmed, fixed

Correct, and worse than described: with `opts.Config == nil` *no* runtime context is rendered at all, so there's no `CLAUDE.md` whatsoever — pre-existing, but it lands squarely on the case where the section matters most. `BuildFromConfig` would also have panicked on a nil config had the guard been relaxed naively, since it dereferences `cfg.Agent`.

`BuildFromConfig` now treats nil as an empty config, the render guard is `opts.Config != nil || scopeAnthropicKeyToShell`, and the `- Agent:` line is omitted when unknown rather than rendering blank. Scoped deliberately: a bare run with no config and no shell-scoping still renders nothing, as before.

## Tests

Companion cases both ways for each: `BASH_ENV` not in the staging mount / staged path matches the filename; nil config renders the section / bare nil config renders a minimal context without inventing one.

`make lint` clean. `internal/deps` has a pre-existing failure in `TestRegistryGithubBinaryURLsExist` — many GitHub release URLs 401 — but that's an artifact of running inside a moat container where the `github` grant injects auth into those requests; CI's `Test` job passes, and `-skip TestRegistryGithubBinaryURLsExist` is green locally.

